PR #323, opened 2026-06-23
fix(reports): quiet the tolerant STATE.json read path (no per-poll log flood)

hyoshi:
The read-only Reports view tries a strict STATE.json parse, then re-reads tolerantly (skipping nonconforming entries). That **works** — the dashboard renders — but it logged a full traceback + one WARNING per skipped entry on **every dashboard poll**. On an account with many legacy / hand-authored campaigns (schema `name`/`id` instead of `campaign_name`/`campaign_id`), the daemon log floods with a traceback + dozens of warnings each refresh, reading as a failure when it is expected graceful degradation.

## Change
Lower the per-render tolerant-path logging WARNING → DEBUG:
- `web/reports.py` `_read_state_safe` (strict-fail → tolerant-retry trace)
- `context/state.py` `_parse_campaigns` / `_parse_action_log` (per-entry skips)

Genuinely-bad paths stay loud (no-path warning, tolerant-also-failed exception, structural errors still raise). Regression tests pin DEBUG-not-WARNING at parse + `build_report_summary` level.

No functional change — the view already rendered; this only quiets the log. python-reviewer: APPROVE. Targets 0.10.13.

https://claude.ai/code/session_01NxBrN8Qn53Tpivt9SumdGn

### DIFF
--- a/mureo/context/state.py
+++ b/mureo/context/state.py
@@ -52,7 +52,10 @@ def _parse_campaigns(
         try:
             parsed.append(_parse_campaign(c))
         except (ValueError, KeyError, TypeError) as exc:
-            logger.warning("skipping unparseable campaign entry: %s", exc)
+            # DEBUG, not WARNING: the read-only Reports view re-parses on every
+            # poll, so a per-entry WARNING would flood the log for a STATE.json
+            # with many nonconforming (e.g. legacy / hand-authored) campaigns.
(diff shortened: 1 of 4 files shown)


hyoshi:
Superseded by #329, which folds in this log-quieting change alongside the missing-account_id tolerance fix. Closing; branch removed.

PR #374, opened 2026-07-08
fix(web): list name-prefixed mureo-* packages in About to match update-check (#365)

### LINKED ISSUE #365: bug(web): Installed-packages list and update-check use inconsistent criteria for mureo.* plugins
## Describe the bug

The "About mureo" web UI shows two related but inconsistent views of third-party `mureo-*` packages:

- **Installed packages** (`GET /api/about`, `mureo/web/about.py`)
- **Update available** banner (`mureo/web/version_check.py` → `mureo/cli/upgrade_cmd.py`)

These two features scan for packages using **different criteria**, so a package can be flagged as "update available" while never appearing in "Installed packages" at all.

## Root cause

`collect_about_info()` (`mureo/web/about.py:90-125`) only considers distributions that register at least one entry point in `ABOUT_ENTRY_POINT_GROUPS` (`mureo/web/about.py:34-38`):

```python
ABOUT_ENTRY_POINT_GROUPS: Final[tuple[str, ...]] = (
    PROVIDERS_ENTRY_POINT_GROUP,           # "mureo.providers"
    RUNTIME_CONTEXT_FACTORY_ENTRY_POINT_GROUP,
    WEB_EXTENSIONS_ENTRY_POINT_GROUP,
)
```

Meanwhile, `_discover_all_mureo_packages()` (`mureo/cli/upgrade_cmd.py:102-128`), used by the update-check path, walks **every** installed distribution via `importlib.metadata.distributions()` and includes anything whose PEP 503-canonicalised name is `mureo` or `mureo-<rest>` (`_is_mureo_package`, `mureo/cli/upgrade_cmd.py:75-79`) — with no entry-point requirement at all.

A package that only registers a `mureo.skills` entry point (no `mureo.providers` / `mureo.runtime_context_factory` / `mureo.web_extensions`) therefore:

- **is** picked up by the update checker (name-prefix match), so "update available" fires when a newer version is published, but
- **is not** picked up by `collect_about_info()` (no matching entry-point group), so it never shows up in "Installed packages".

## To reproduce

1. `pip install` a package named `mureo-<something>` that only declares `[project.entry-points."mureo.skills"]` in its `pyproject.toml` (no `mureo.providers` / `mureo.runtime_context_factory` / `mureo.web_extensions`).
2. Open the configure UI → About tab. The package does not appear under "Installed packages".
3. Publish a newer version of that package to the configured package index.
4. Reopen the UI (or trigger the update check). The "Update available" banner lists the package — despite step 2 showing it as not installed.

## Expected behavior

The two features should agree: any package the updater considers "an installed mureo package" should also be visible in the Installed packages list (or, alternatively, the updater should scope itself to the same entry-point-registered set that About uses). Currently they diverge because one keys off entry-point registration and the other off name-prefix matching against all installed distributions.

## Environment

- **mureo version**: main branch as of 2026-07-08
- **Installation method**: pipx / pip install into the same venv as `mureo`
- **Affected package (reproduction case)**: `mureo-logly-tools` (private, LOGLY-internal — registers only `mureo.skills`, no `mureo.providers`)

## Component

- [x] Web UI (About / configure)
- [x] CLI (`mureo upgrade`)
- [ ] MCP server
- [ ] Google Ads client
- [ ] Meta Ads client
- [ ] Auth / OAuth
- [ ] Other

## Additional context

We looked into whether `mureo.providers`' entry point is safe to declare purely to become visible in the Installed list, without implementing a "real" provider. Per `mureo/core/providers/base.py`'s `BaseProvider` Protocol, only three class attributes (`name`, `display_name`, `capabilities`) are required — no methods — and `mureo/web/about.py` never calls `.load()` on the entry point, so a metadata-only provider declaration is safe and doesn't risk crashing discovery or MCP tool collection (`mureo/core/providers/registry.py`, `mureo/mcp/tool_provider.py` both isolate per-entry-point failures and gate real usage behind `isinstance` checks).

We plan to work around this on our side by registering a lightweight `mureo.providers` entry point for `mureo-logly-tools` (tracked in logly/mureo-logly-tools). Filing this issue so the underlying inconsistency in `mureo` itself is tracked and can be fixed at the source (e.g. by adding a `mureo.skills`-aware group to `ABOUT_ENTRY_POINT_GROUPS`, or by having the updater require entry-point registration too) for other `mureo.skills`-only plugins that hit the same gap.

hyoshi:
## Summary

Fixes #365 — the "About mureo" Installed-packages list and the "Update available" banner used **inconsistent discovery criteria**, so a package could be flagged as "update available" while never appearing in "Installed packages".

- **About** (`collect_about_info`) listed only distributions registering an entry point in the providers / runtime_context_factory / web_extensions groups.
- **Update-check** (`_discover_all_mureo_packages`) flags **any** `mureo` / `mureo-*` distribution by PEP 503 name prefix — no entry-point requirement.

A `mureo.skills`-only plugin (e.g. `mureo-logly-tools`) was therefore reported as updatable yet shown as not installed.

## Fix

`collect_about_info()` now unions **two discovery axes**, deduplicated by PEP 503 canonical name (which is also the display name, matching the update banner's labelling):

1. **Entry-point** — contributors across **every** mureo extension group. Adds `mureo.skills`, `mureo.policy_gates`, `mureo.analytics` to `ABOUT_ENTRY_POINT_GROUPS` (previously only providers / rcf / web_extensions). Catches plugins **not** named `mureo-*` (e.g. a third-party `acme-ads` provider).
2. **Name-prefix** — every installed `mureo` / `mureo-*` distribution via the **same** `_discover_all_mureo_packages()` the updater uses. Closes the reported gap.

This guarantees **updater's set ⊆ Installed list**, and both surfaces now label a package identically (canonical form).

Also hardened `_safe_version`: a missing / `None` / broken-metadata version (e.g. `importlib.metadata.version()` returning `None` for a dist whose `METADATA` has a Name but no Version header — it does **not** raise) degrades to the `unknown` placeholder, so the payload's `{"name": str, "version": str}` shape always holds. The Python-3.10 `dist is None` fallback path in `_resolve_distribution` routes through it too. The About endpoint never 500s.

## Test plan

- [x] `tests/test_web_about.py` — updated `ABOUT_ENTRY_POINT_GROUPS` assertion; added cases: name-prefixed plugin w/o entry points is listed; dedupe across the two axes by canonical name; skills / policy_gates / analytics groups scanned; name-prefix discovery failure isolated; `version()==None` → placeholder (both name-prefix axis and the `dist is None` fallback); mixed-case name deduped + displayed canonical; ⊆ property (every updater package listed exactly once). **34 passed.**
- [x] Related suite (`-k "about or version_check or upgrade or handlers"`) — **517 passed**.
- [x] `ruff` + `black` clean.
- [x] Two rounds of `code-reviewer` review; all WARNING findings addressed.

## Scope

OSS `mureo` only — `mureo/web/about.py` + tests. No change to the updater, MCP server, or ad-platform clients.

### DIFF
--- a/mureo/web/about.py
+++ b/mureo/web/about.py
@@ -1,40 +1,66 @@
 """Version/package payload for the configure UI's "About mureo" tab (#229).
 
-Resolves the installed mureo version plus every distribution that
-contributes to mureo's plugin entry-point groups (providers, runtime-
-context factories, web extensions). Discovery is entry-point based, so
-a freshly installed plugin appears automatically — no per-plugin wiring.
+Resolves the installed mureo version plus every installed mureo plugin,
+discovered along TWO axes that together match what ``mureo upgrade`` /
+the update-check consider an installed mureo package (#365):
+
+1. **Entry-point** — any distribution contributing to one of mureo's
+   plugin extension groups (providers, skills, runtime-context factories,
+   web extensions, policy gates, analytics). Catches a plugin whose
+   distribution is NOT named ``mureo-*`` (e.g. a third-party ``acme-ads``
+   that registers ``mureo.providers``).
+2. **Name-prefix** — every installed ``mureo`` / ``mureo-*`` distribution,
+   using the SAME discovery the updater uses
+   (:func:`mureo.cli.upgrade_cmd._discover_all_mureo_packages`). Catches a
+   plugin that registers only, say, ``mureo.skills`` and would otherwise be
+   flagged "update available" yet never listed here (#365).
+
+The two are unioned and deduplicated by PEP 503 canonical name (which is
+also the display name, matching the update banner's labelling), so the
+"Installed packages" list and the "Update available" banner agree —
+whenever both observe the same installed metadata — about what is
+installed and how it is named.
 
 Only :mod:`importlib.metadata` (stdlib) is consulted. The payload
 carries nothing environment-specific: no secrets, no file paths — only
 distribution names and versions.
 
 Fault isolation mirrors :mod:`mureo.web.extensions`: one broken
-plugin's metadata (a raising ``dist`` lookup, a corrupted group) skips
-just that item, never the endpoint.
+plugin's metadata (a raising ``dist`` lookup, a corrupted group) — or a
+failing name-prefix walk — skips just that item, never the endpoint.
 """
 
 from __future__ import annotations
 
 from importlib.metadata import (
-    PackageNotFoundError,
     entry_points,
     packages_distributions,
     version,
 )
 from typing import Any, Final
 
-from mureo.core.providers.registry import PROVIDERS_ENTRY_POINT_GROUP
+from mureo.analytics.registry import ANALYTICS_ENTRY_POINT_GROUP
+from mureo.cli.upgrade_cmd import _canonicalise, _discover_all_mureo_packages
+from mureo.core.policy import POLICY_GATES_ENTRY_POINT_GROUP
+from mureo.core.providers.registry import (
+    PROVIDERS_ENTRY_POINT_GROUP,
+    SKILLS_ENTRY_POINT_GROUP,
+)
 from mureo.core.runtime_context import RUNTIME_CONTEXT_FACTORY_ENTRY_POINT_GROUP
 from mureo.web.extensions import WEB_EXTENSIONS_ENTRY_POINT_GROUP
 
 #: Entry-point groups scanned for contributing distributions. Reuses the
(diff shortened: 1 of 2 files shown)


hyoshi:
Superseded — these exact changes were merged into main via #375 (commit b48a03f). The SC-tenant-scope branch was accidentally based on this branch, so merging #375 also merged this fix. `git diff main` for about.py/test_web_about.py is empty. Closing as redundant; issue #365 is resolved by b48a03f.

PR #412, opened 2026-07-14
fix(policy): enforce budget guardrails on the provider ABI's own budget key

hyoshi:
## Problem

`StrategyPolicyGate` (#360) hard-enforces the operator's STRATEGY.md `## Guardrails` budget caps before dispatch — but it finds the proposed budget by matching a fixed set of argument keys:

```python
_BUDGET_KEYS = ("daily_budget", "proposed_daily_budget", "amount")
...
for micros_key in ("budget_amount_micros", "amount_micros"):
```

Those are the spellings the first-party `google_ads_*` / `meta_ads_*` tools use. **`daily_budget_micros` is not among them** — yet that is the provider ABI's *own* budget field (`Campaign.daily_budget_micros`, `CreateCampaignRequest.daily_budget_micros`, `mureo/core/providers/models.py`) and the name `docs/plugin-authoring.md` explicitly tells plugin authors to use:

> integer micros … on every monetary field (`daily_budget_micros`, `cost_micros`, `cpc_bid_micros`)

The gate *does* run on plugin tools — `_evaluate_policy_gates` has no first-party allow-list — so the result is that **a provider plugin which follows the ABI has every budget rule silently skipped**. `_proposed_budget()` returns `None`, and `max_daily_budget_per_campaign`, `max_daily_budget_increase_pct` and `max_total_daily_budget` are all bypassed.

An operator's cap was enforced on Google and Meta while that plugin's budget writes went straight through — no error, no warning, just an unenforced cap on a real-spend surface. The gate's key list was a correct-but-incomplete union of *platform-native* names; it was never reconciled against the *ABI* name.

## Fix

Read `daily_budget_micros` alongside the other micros keys.

Deliberately an **explicit key**, not a generic `*_micros` suffix rule: a suffix rule would also match `cpc_bid_micros` / `cost_micros`, and treating a bid as a budget would misclassify bid-only mutations.

## Impact

- Any ABI-faithful provider plugin now has the budget caps applied to its writes.
- **First-party behaviour is unchanged.** No `_tools_*.py` inputSchema declares `daily_budget_micros` (the Google budget tools use `amount` / `amount_micros` / `total_amount*`; Meta uses `daily_budget` / `lifetime_budget`). `_BUDGET_KEYS` is still checked before the micros keys, so precedence is untouched.
- Fail-open contract is unchanged: no `## Guardrails` section ⇒ abstain.

## Tests

Three added to `tests/test_strategy_gate.py` (absolute cap, under-cap allow, % increase). Two of them fail without the one-line fix — verified by reverting it:

```
FAILED test_canonical_abi_daily_budget_micros_denies
FAILED test_canonical_abi_daily_budget_micros_respects_increase_pct
```

Full suite: 28 passed. ruff / black / mypy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **The budget guardrail now fires on provider plugins.** `StrategyPolicyGate`
+  matched a proposed budget against a fixed set of argument keys
+  (`daily_budget`, `amount`, `amount_micros`, …) — the spellings the
+  first-party `google_ads_*` / `meta_ads_*` tools use. It did
+  not match `daily_budget_micros`, which is the provider ABI's *own* budget
+  field (`CreateCampaignRequest.daily_budget_micros`) and the name
+  `docs/plugin-authoring.md` tells plugin authors to use. A provider plugin
+  that followed the ABI therefore had **every** budget rule silently skipped:
+  an operator's `max_daily_budget_per_campaign` was enforced on Google and Meta
+  while that plugin's budget writes went straight through — no error, no
+  warning, just an unenforced cap on a real-spend surface. `daily_budget_micros`
+  is now read alongside the other micros keys, so the caps
+  (`max_daily_budget_per_campaign`, `max_daily_budget_increase_pct`,
+  `max_total_daily_budget`) apply to any ABI-faithful plugin. Behaviour for
+  first-party tools is unchanged.
+
 ## [0.10.23] - 2026-07-13
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


hyoshi:
Superseded by #416 — close します。

この PR は built-in の key scan に `daily_budget_micros`（provider ABI の canonical な予算フィールド）を足す案でした。#414 への回答として **#416 が宣言 seam を採用**し、「plugin は自分の予算キーを宣言する / built-in scan は Google/Meta の綴りのまま」という設計が確定したため、この変更は方針と競合します（実際 main と conflict しています）。

さらに #417 / #421 で、**宣言が届かない passthrough plugin は normalize-and-delegate（自前 PolicyGate）を使う**ことが正式な答えとして文書化されました。ABI に忠実な plugin は宣言、raw body に予算を持つ plugin は自前 gate、という住み分けです。

この PR が報告した「ABI に忠実な plugin の予算 cap が無言で素通りする」問題自体は、#416（宣言 seam）と #418（宣言が増加率 / total cap を無効化しないよう修正）で解消済みです。

Issue #545, opened 2026-08-06
Import externally-made changes (platform UI edits) into action_log, so mureo's guarantees survive manual operation


hyoshi:
## Summary

mureo has no way to learn about changes made outside itself. When an operator works in a platform's own UI — which is normal professional behaviour, not misuse — the change is never dispatched through mureo, so `StrategyPolicyGate` never evaluates it, it is never promoted to `action_log`, no `observation_due` window is attached, and `/daily-check`'s evidence step has nothing to review.

The result is that mureo's core guarantees silently do not apply to a large and unpredictable share of real operations, and mureo cannot tell the difference between "nothing happened" and "something happened that I cannot see".

This is a platform-agnostic gap. It should be solved once in core, with a thin per-platform adapter, and an ABI hook so bridge and plugin platforms can participate.

## Motivating incident (anonymised)

An operator tightened Display placement exclusions over several UI sessions. Delivery on two campaigns collapsed to zero and stayed there for about a week. Because none of the work was in `action_log`, there was no record connecting "delivery died" to "exclusions were added", so recovery proceeded by guesswork: bulk-revert (no effect), then a bid-ceiling change that reset the bidding strategy into a learning phase and made things worse, then a full rebuild that was probably unnecessary.

Roughly 1.5 days were lost to detection and 6 days to failed diagnosis. The diagnosis loss is directly attributable to this gap.

## Why polling, not backfill

Google Ads `change_event` returns at most 100 rows with no pagination. During the post-mortem for the incident above, the investigation hit exactly this wall: a single bulk exclusion edit consumed the entire window, making everything before it permanently unreachable.

**History cannot be reconstructed after the fact.** It must be captured continuously or it is lost. mureo already ships an always-on daemon (`mureo service install`), which is the natural home for this.

## Cross-platform scope

| Platform | Change-history source | Available in mureo today |
|---|---|---|
| Google Ads | `change_event` (100-row cap, no paging) | read-only tool exists (`google_ads_change_history_list`), not imported into `action_log` |
| Meta Ads | Ad Account Activity (`/activities`) | no tool |
| Yahoo Ads (Search / Display) | vendor change history | no tool |
| LINE Ads | vendor API | no tool |
| SmartNews Ads | vendor API | no tool |
| Amazon Ads (official-MCP bridge) | vendor API | no tool |
| TikTok Ads (hosted connector) | connector's own tools | out of mureo's data path entirely — this gap is worst here |
[...]

PR #569, opened 2026-08-07
Treat a bulk operation as one revertible unit in action_log and rollback

### LINKED ISSUE #549: Treat a bulk operation as one revertible unit in action_log and rollback
## Summary

mureo has rollback (`rollback_plan_get`, `rollback_apply`, `mureo rollback list/show`), but it reasons about individual allow-listed operations. There is no notion of **a batch as one revertible unit**, so "undo what I did on Monday" is not expressible — the operator has to reconstruct the member entities by hand.

After a bulk exclusion pass killed delivery, the revert attempt was manual: the operator had to work out which exclusions had been added across several sessions and remove them by hand, ending up deleting 80+ entries in one pass without confidence that the set matched what had been added. It did not restore delivery, and because the revert itself was unverifiable, it could not be ruled out as incomplete — which contributed to the decision to rebuild the campaigns instead.

An unverifiable revert is nearly as bad as no revert: it leaves the operator unable to eliminate their own fix as a variable.

## Proposed shape

1. Group operations dispatched as one logical batch under a shared batch id in `action_log`, so a multi-entity change (N exclusions, N keywords, N ad status changes) is one reviewable and reversible unit.
2. Let `rollback_plan_get` accept a batch id and produce a plan covering every member, reporting explicitly which members are reversible and which are not, before anything is applied.
3. Report partial reversibility honestly. A batch where 60 of 80 members can be restored must say so rather than reporting success — the operator's next diagnostic step depends on knowing whether the revert was complete.

## Cross-platform scope

Batch grouping is a core `action_log` concern and should be platform-agnostic. What differs is how far reversal can actually go:

- **Native platforms** (Google Ads, Meta Ads): full reversal for allow-listed operations, as today.
- **Bridge / plugin platforms** (`plugin:<dist>`): today a reversal hint is *recorded for visibility but not executed*, and per-call reversal for entity-scoped plugin mutations is a known open problem. Batch grouping is still valuable here even without execution — it turns "figure out what you did" into "here is the exact list", which is most of the work.
- **Hosted connectors** (`tiktok_ads`): auto-rollback is unavailable by design since the call never passes through mureo. Batch grouping should still apply to entries recorded via `mureo_state_action_log_append`, so the manual revert has an accurate checklist.

In all three cases the deliverable is the same: the operator should never have to reconstruct from memory what a change set contained.

hyoshi:
Closes #549.

## Why

Rollback existed but reasoned about **one** allow-listed operation at a time, so "undo what I did on Monday" was not expressible — after a bulk pass the operator had to reconstruct the member entities by hand. The incident behind this issue ended with 80+ entries deleted in one pass with no confidence the set matched what had been added, and because the revert itself could not be verified, the operator could not rule their own fix out as a variable.

### 1. The boundary is declared, not guessed

A bulk pass is many tool calls and nothing in a single call says which others belong with it. Inferring membership (same minute, same campaign, same tool) would be a heuristic, and a heuristic that silently omits a member re-creates exactly the failure this is meant to prevent. So:

- `mureo_batch_begin` (takes a `label`) / `mureo_batch_end` (returns the exact `member_indices` + `platforms`) / `mureo_batch_status`.
- Nesting is refused; closing with none open is refused. Neither is resolved silently.
- Batch state lives in STATE.json (`batches`), not process memory, so a host restarting the MCP server mid-pass does not silently stop collecting. Records are kept after close (`ended_at`) so a `batch_id` still resolves to the operator's own label later.

### 2. Membership is stamped in core, at one choke point

Every recording path — native status toggles (`native_reversal`), `mureo_state_action_log_append`, and bridged/plugin promotion (`plugin_semantics.record_mutation_action_log`) — already converges on `append_action_log`. The stamp is applied **there**, inside the state lock.

That is the whole cross-platform story: no per-platform code, no plugin ABI hook, no tool-schema change. Threading a `batch_id` through tool arguments would have worked for native tools only — a bridged tool's input schema belongs to the bridged platform, not to mureo, and many declare `additionalProperties: false`.

Reversals appended by `rollback_apply` are explicitly excluded (`join_active_batch=False`); otherwise reverting a batch would grow the batch it reverts.

### 3. `rollback_plan_get` plans a whole batch, gaps included

Pass `batch_id` instead of `index` and the response covers **every** member:

- `coverage` — `full` / `partial` / `none` / `empty`
- `platform_coverage` — the same verdict **per platform key**, because reversibility is not uniform across platforms
- `counts` — reversible / reversible_with_caveats / irreversible / nothing_to_reverse / already_reversed / total
- `members[]` — each with its `index`, `platform`, `reversibility` and, when it is a gap, the **reason**
- `apply_order` — the reversible indices, newest first

Each member is classified by the existing `plan_rollback` allow-list, so grouping loosens no guarantee. A batch where 60 of 80 members can be restored reports `partial` with the other 20 named and explained, **before** anything is applied — planning is pure and writes nothing.

## Honest limits (stated in the docs, not smoothed over)

| Platform kind | Joins a batch | Reversal |
|---|---|---|
| Native Google / Meta | Status toggles automatically; **every other mutation** (budget, keywords, exclusions) only if the agent records it with `mureo_state_action_log_append` | Executed for allow-listed operations |
| Bridged / plugin (`plugin:<dist>:<provider>`, e.g. Amazon Ads) | Automatically, on successful mutation | Executed only when the reversal names a *registered* plugin tool; otherwise reported `irreversible` with the reason |
| Hosted connectors (`tiktok_ads`) | Via the agent's own `mureo_state_action_log_append` — mureo is not in the data path | Never executed by mureo. The plan is an accurate manual checklist |
| Search Console | **No** — its mutations are not recorded in `action_log` at all | n/a |

## Compatibility

`ActionLogEntry.batch_id` and `StateDocument.batches` are both optional and emitted **only when present**, so an existing STATE.json parses unchanged and gains no new key on the next write (pinned by a round-trip test). `batch_id` is appended last on the dataclass, preserving positional-constructor compatibility. `rollback_plan_get`'s `index` form is unchanged; `batch_id` is an additive alternative (`oneOf`).

## Tests

New `tests/test_batch_revertible_unit.py` (21 tests), deliberately multi-platform: **google_ads** and **meta_ads** (native), **tiktok_ads** (hosted connector) and **plugin:mureo-amazon-ads-bridge:amazon_ads** (bridged), including a four-platform batch where each member enters `action_log` through a different recording path, and a mixed-reversibility batch spanning three platforms.

Not vacuous — two injected regressions, both caught:

- dropping `batch_id` propagation in `stamp_batch` → **8 failures**
- reporting `NOT_SUPPORTED` members as `REVERSIBLE` → **3 failures**, including `test_partial_reversibility_is_reported_before_anything_is_applied` and `test_coverage_is_reported_per_platform`

## Not in this PR

`rollback_apply` still takes one `index` per call; applying a batch is a loop over `apply_order`. A single "apply the whole batch" result code would have to summarize partial failure into one status, which is the reporting problem this issue exists to remove. No CLI surface for batches — `mureo rollback list` / `show` remain entry-by-entry, noted in `docs/cli.md`.

## Docs

`docs/mcp-server.md` (tool count 205 → 208, new Batch section + batch-plan fields + the per-platform table), `docs/strategy-context.md` (`batches` + `batch_id` schema), `docs/architecture.md`, `docs/cli.md`, `AGENTS.md`, `skills/_mureo-shared/SKILL.md` + `skills/search-term-cleanup/SKILL.md` (and their `mureo/_data/skills/` mirrors), CHANGELOG.

## Gates

`ruff check .` / `black --check .` / `mypy mureo --ignore-missing-imports` clean. `python -m pytest`: 7795 passed, 12 failed — the identical 12 that fail on `origin/main` in this environment (9 × `test_live_clients.py`, plus 3 tool-registry tests that count locally-installed plugin tools). No new failures.


hyoshi:
Review fixes pushed in 8f2681b. All three MEDIUMs addressed; the pre-existing `is_read_only_tool_name` bug is deferred to a follow-up PR — measurement below.

## 1. Membership is now tamper-proof

An explicit `batch_id` is validated, not trusted: it must name a batch that was **declared** and is **still open**. The check lives in `append_action_log`, inside the state lock, so no caller — MCP handler, library user, or a future recorder — can bypass it.

**Closed batches are refused, deliberately.** `mureo_batch_end` reports a `member_count` the operator keeps; if a later append could still join, that number silently stops being true. Closing is final. For the backfill/import case (#545), the affordance is to **declare a batch for the import** — which already works and gives the imported set an honest label and start time, rather than retrofitting someone else's.

`batch_id` on `mureo_state_action_log_append` is therefore no longer documented as "import/backfill"; it is an explicit **assertion** that this entry belongs to the batch you believe is open, and it fails loudly when it does not.

The reviewer's exact reproduction is now a test — no `begin`, arbitrary id, append refused, and `rollback_plan_get` for that id reports `coverage: "empty"` with no members.

## 2. A forgotten `end` announces itself

Both halves, and I'll say why. `mureo_batch_status` carries a `warning` for the caller who asks — but the caller who **forgot** the batch is open is by definition not asking, and every mutation dispatched meanwhile adds another member. So the thorough half is the one that matters: the warning is appended to the result of every **mutating** tool call (built-in and plugin), in the same soft-enforcement shape as `_maybe_append_strategy_reminder`.

Re-emitted per mutation rather than latched once per process — each mutation is a *new* member wrongly joining, not a repeat of the old problem. Reads do not warn (they add no members). Threshold 24h: a bulk pass is one working session, so a batch open across a day boundary has almost certainly been forgotten. Opt out with `MUREO_DISABLE_BATCH_REMINDER=1`.

**Nothing is auto-closed**, per the review instruction — and the code says why: an auto-close would trade a visible wrong answer for an invisible one, since entries after it would silently stop joining and no one would be told. An unparseable `started_at` returns `None` rather than "fresh" — an unknown age must not pass for a small one.

## 3. `_resolve_path` lifted

Now `mureo/mcp/_helpers.py::resolve_workspace_path`, imported by name in both `_handlers_mureo_context` (9 call sites) and `_handlers_batch`. The private cross-module import is gone; the module docstring records that it is a sandbox boundary and therefore single-sourced. No test referenced the old name.

## Pre-existing `is_read_only_tool_name` — **follow-up PR**, and it is not the one-liner it looks like

Measured before deciding:

- **Native**: 70 of 208 tool names flip under a trailing-verb rule. None contains a write verb, so the native direction is safe.
- **Plugin surface installed here (294 tools)**: 23 flip, and **13 of those are classified `mutating`**.

That second number is the blocker. `server._register_pattern_fallbacks` skips `register_pattern_fallback_tool(name)` when `is_read_only_tool_name(name)` — so those 13 would be **newly exempted from the guardrail money pattern-scan**. Among them:

```
reporting-delete_report              mutating   <- a DELETE that would read as a read
reporting-create_campaign_report     mutating
reporting-create_report              mutating
logly_ads_context_merge_adgroup_list mutating
amc-execute_query                    mutating
```

A naive trailing-verb rule is therefore **wrong, not merely broad**: it silently weakens a denial heuristic on real installed tools. The correct fix has to be write-verb-aware (a segment ending `_report` that also carries `delete`/`create` is not a read) and needs its own tests across `test_strategy_gate_pattern_fallback.py`, `test_mcp_plugin_semantics.py`, `test_rollback.py` and `test_cli_rollback.py`. Separate PR, opening immediately after this one — not an issue, not a note.

Meanwhile the batch test requested in review is in, covering **both** spellings:

- `campaign_management-list_campaigns` (bridged) → `nothing_to_reverse` — correct today
- `google_ads_campaigns_list` (native) → `irreversible` — **misclassified today**, pinned with a comment naming the defect and the assertion the follow-up flips
- `apply_order` excludes both either way, so no read is ever offered for reversal

## Gates

`ruff` / `black --check` / `mypy` clean. `pytest`: **7806 passed, 12 failed** — the identical 12 that fail on `origin/main` here. No new failures.

Regression-injected the three new protections, all caught:

| Injection | Result |
|---|---|
| `ensure_joinable` accepts any id | 2 failed |
| closed batches rejoinable | 1 failed |
| `stale_batch_warning` always `None` | 4 failed |

hyoshi:
Cleanups pushed in 4d88064.

## MEDIUM — batch staleness now goes through the clock seam

`batch_open_hours` defaults to `clock.server_now()`. You were right that the production path was the only caller outside the seam, and that no test would have noticed the drift — so there is one now: `test_staleness_follows_the_server_clock_seam` freezes `mureo.core.clock.server_now` and asserts the verdict follows it. Reverting to `datetime.now(timezone.utc)` fails that test and nothing else (verified by injection).

One deviation from the snippet you gave, forced by the import graph. A module-level `from mureo.core import clock` in `mureo/context/batch.py` does not work:

```
mureo.core.__init__ -> runtime_context -> state_store
                    -> mureo.context.state -> mureo.context.batch
ImportError: cannot import name 'BatchError' from partially initialized module
             'mureo.context.batch' (most likely due to a circular import)
```

Verified rather than assumed — `mureo.core.clock` itself imports only `datetime`, but reaching it executes `mureo/core/__init__.py`, which closes the cycle. So the import is lazy, inside the `now is None` branch, and still resolved through the **module** (`clock.server_now()`) rather than a bound name, which is what keeps `monkeypatch.setattr(mureo.core.clock, "server_now", …)` effective. Both the reason and the verification are in the docstring.

No `datetime.now` remains in the module.

## LOW — stale `_resolve_path` references

Retargeted at `mureo/context/batch.py`, `mureo/context/conversion_overrides.py` and `tests/test_mcp_tools_mureo_context.py`.

Left alone deliberately: `mureo/policy/declarations._resolve_path` and the reference to it in `mureo/amazon_ads/money_paths.py`. That is an unrelated function of the same name (argument-path resolution for policy declarations), not the workspace sandbox helper.

## LOW — deferred-fix test is now self-sufficient

`test_native_read_only_member_is_not_counted_as_a_gap` no longer says "lands in its own PR" with nothing to follow. It now records, in the docstring:

- **the defect**, with the failing call — `is_read_only_tool_name("google_ads_campaigns_list") -> False`
- **why the obvious fix is wrong** — `_register_pattern_fallbacks` skips the guardrail money pattern-scan for anything reading as a read, and all **13** mutating plugin tools a naive suffix rule would strip are listed by name:

```
amc-execute_query
logly_ads_context_merge_adgroup_list
reporting-create_campaign_report
reporting-create_inventory_report
reporting-create_product_report
reporting-create_report
reporting-delete_report          <- a DELETE reading as a read
yahoo_ads_create_placement_url_list
yahoo_ads_display_create_placement_url_list
yahoo_ads_display_remove_placement_url_list
yahoo_ads_display_update_placement_url_list
yahoo_ads_remove_placement_url_list
yahoo_ads_update_placement_url_list
```

- **what a correct fix must do** — match a trailing verb only when no write verb appears elsewhere in the same segment
- **which test files it touches** — `test_strategy_gate_pattern_fallback.py`, `test_mcp_plugin_semantics.py`, `test_rollback.py`, `test_cli_rollback.py`
- **the single assertion that flips** — `by_index[0]` becomes `NOTHING_TO_REVERSE`; the other two and `apply_order` are unchanged

Someone finding this in six months can act on it without this thread.

Thanks for reproducing the measurement independently — your 26 vs my 23 total flips is the interesting difference, but since the 13 mutating ones match exactly, the conclusion holds either way.

## Gates

`ruff` / `black --check` / `mypy` clean. `pytest`: **7807 passed, 12 failed** — the identical 12 that fail on `origin/main` here. No new failures.

hyoshi:
Fixed in 2d5d5c8. The bug is mine and the diagnosis is right — thanks for tracing it back here rather than patching it in #545.

## The fix

`stamp_batch` now uses `dataclasses.replace(entry, batch_id=batch.batch_id)`. Fixed as a class of bug rather than an instance: every field crosses by construction, so no future field can opt out of batching by omission.

The docstring records why, including the concrete cost — that an enumerated rebuild dropped `origin` / `external_id` and with them the `is_external` marker that stops a forged `reversible_params` on an imported entry from planning as a real reversal.

## Audit of the other `ActionLogEntry(` sites

Seven in the tree. Six are not this defect:

| Site | Verdict |
|---|---|
| `native_reversal.py:222` | New entry from scratch — nothing to carry over |
| `plugin_semantics.py:416` | Same |
| `tools_creative_studio.py:505`, `:738` | Same |
| `rollback/executor.py:194` | Same |
| `_handlers_mureo_context.py:289` | Builds from an untrusted MCP dict; enumerating is *correct* here — it is the whitelist of what a caller may set |
| `state_codec.py:291` | **Enumerates, and must** — it maps to an external JSON schema, so `replace` cannot help. Its omission is silent in exactly the same way, so the round-trip test below covers it |

`stamp_batch` was the only rebuild-of-an-existing-entry.

## Tests

Both driven off `dataclasses.fields(ActionLogEntry)`, not a hand-written subset:

- `test_stamp_batch_changes_only_batch_id` — the pure function, so a failure localizes here rather than in the codec.
- `test_every_field_survives_the_append_round_trip` — through `stamp_batch`, the codec, onto disk and back.

Verified against three separate failure modes by injecting a simulated #545 `origin` field:

| Injection | Result |
|---|---|
| simulated `origin` + the old enumerating `stamp_batch` | **both tests fail** (this is the reported bug) |
| simulated `origin`, `stamp_batch` fixed, codec unaware of it | **round-trip test fails** — so the codec half is genuinely covered |
| simulated `origin` on the dataclass, absent from the test's value map | **both fail** with `ActionLogEntry gained field(s) ['origin'] with no value in _ENTRY_FIELD_VALUES…` |

That third one is the anti-rot property: the test cannot silently stop covering a field the way the code silently stopped copying one.

## One observation, not acted on

The same shape is live one level up. `StateDocument` is rebuilt field-by-field in five `mureo/context/state.py` mutators (`append_action_log`, `upsert_campaign`, `set_report`, `set_platform_metrics`, `set_conversion_action_types`). That is pre-existing, but this PR is direct evidence it bites: adding `batches` required hand-threading it into all five, and missing one would have silently closed an open batch on the next campaign upsert.

`end_batch` and `begin_batch` already use `replace`. Converting the other five is mechanical but touches merge semantics with real comments attached, and you said this was the only change — so I have not. Say the word and I will, in its own PR, with a field-list-driven preservation test per mutator.

## Gates

`ruff` / `black --check` / `mypy` clean. `pytest`: **7809 passed, 12 failed** — the identical 12 that fail on `origin/main` here.

One note on that number: an earlier run showed 13, with `test_web_handlers.py::TestServeAbout::test_about_unmocked_lists_mureo` added. It is not related — that test binds a real local HTTP server with a 2s `urlopen` timeout, and a sibling worktree was running its own full suite at the same time. It passes in isolation and did not recur on a clean re-run. Flagging rather than quietly reporting the good number.

PR #570, opened 2026-08-07
Validate tracking-parameter consistency on ad create/update and in /tracking-health

### LINKED ISSUE #550: Validate tracking-parameter consistency on ad create/update and in /tracking-health
## Summary

mureo does not validate that an ad's final URL tracking parameters are consistent with the campaign the ad actually lives in. Nothing catches an ad shipped into campaign B carrying campaign A's tracking tags.

This class of defect is more dangerous than a delivery fault because it is silent: delivery looks healthy, spend looks healthy, and the analytics everyone downstream trusts is quietly wrong. Nobody investigates, because nothing appears broken.

An account runs two audience-segment Display campaigns against the same eight landing pages. The segments are distinguishable in analytics *only* by a `utm_campaign` prefix — one segment uses `<segA>01..08`, the other `<segB>01..08`, where the number identifies the article and the prefix identifies the segment.

During setup, 16 ads were uploaded into segment B's campaign carrying **segment A's prefix**. The same account also had ads uploaded into a Display campaign carrying a third scheme belonging to a social campaign entirely (`utm_source=sns_cp&utm_medium=cpci`).

**These ads never served** — a per-ad report confirms 0 impressions, 0 clicks and 0 cost on every one of them, and they were later removed. So no data was actually corrupted this time.

That is precisely why it is worth filing. The misconfiguration reached the platform and was caught by a human noticing and re-uploading, not by any check. Had those ads been left enabled, a month of segment-level reporting would have been silently unusable, and the error would only have surfaced — if ever — during an unrelated audit. In this case it *was* found during an unrelated audit, weeks later, by which time the only reason it did not matter was luck.

The failure mode is copy-paste during multi-campaign setup, which is routine and will recur.

## Proposed shape

A pre-flight validation on ad create/update, plus a read-only audit that can run across an account:

1. **Intra-account consistency.** Flag when the same campaign-identifying tracking value appears under more than one campaign, or when ads within one campaign disagree with each other. This is detectable with zero configuration and would have caught the incident above at upload time.
2. **Convention conformance (opt-in).** Let a naming convention be declared in `STRATEGY.md` (e.g. `utm_campaign` must carry a token derived from the campaign name) and validate against it. Optional — accounts without a declared convention still benefit from step 1.
3. **Required-parameter presence.** Flag missing `utm_source` / `utm_medium` / `utm_campaign` and untagged final URLs, where the account uses tagging at all.

Severity should reflect delivery state: a mis-tagged ad that has already served is a data-integrity incident and needs a reporting caveat; a mis-tagged ad that has not yet served is a cheap fix. The check should say which it is.

Placement: `/tracking-health` is the natural home for the account-wide audit; the ad-creation path is the natural home for the pre-flight. Creative upload skills should route through it the same way creatives already route through `/ad-policy-check`.

## Cross-platform scope

Fully platform-agnostic — every platform mureo drives has ads with destination URLs, and the copy-paste failure mode is identical everywhere:

- **Native**: Google Ads (`final_urls` on RSA / RDA / image ads), Meta Ads (link URL and `url_tags` on creatives).
- **Bridges**: Yahoo, LINE, SmartNews, LOGLY, Amazon — all expose ad creation with destination URLs.
- **Hosted connectors**: validation can run best-effort on whatever the connector reports.

The check operates on strings already present in each platform's ad records, so a shared core validator plus a small per-platform accessor ("give me this ad's destination URLs") covers every platform including plugins, via the provider ABI.

hyoshi:
Closes #550

## The problem

Nothing caught an ad shipped into campaign B carrying campaign A's tracking tags. That defect is more dangerous than a delivery fault because it is **silent**: delivery looks healthy, spend looks healthy, and the analytics everyone downstream trusts is quietly wrong. Nobody investigates, because nothing appears broken.

## The design decision: where "inconsistent" comes from

The hard part is not detecting a difference — it is not inventing one. mureo does not know what a correct `utm_campaign` looks like for an account: a prefix that identifies an audience segment in one account is a campaign month in the next. A check that guessed the convention would produce false positives, and a check that produces false positives gets muted — at which point it detects nothing at all.

So the zero-configuration checks derive their verdict **entirely from evidence already in the account**, and operator intent — the one thing evidence cannot supply — is **declared, never inferred**.

| Code | Fires when | Config |
| --- | --- | --- |
| `foreign_campaign_scheme` | Some ads of campaign X carry a whole campaign-identifying signature that is the **sole** signature of **exactly one** other campaign Y (≥2 ads), while X also contains a different signature. | none |
| `same_destination_scheme_conflict` | Two ads in one campaign send the **same** landing page under different identifying signatures. Needs no second campaign, so it still fires when the source campaign is out of scope. | none |
| `missing_tracking_parameter` / `untagged_final_url` | An ad lacks what **every** other tagged ad in its own campaign carries. mureo does not assert that every account must carry `utm_source`/`utm_medium`/`utm_campaign`. | none |
| `missing_required_parameter` / `convention_violation` | The account declared a `## Tracking Convention` in STRATEGY.md and an ad breaks it. | opt-in |

Neither scheme check declares which group is *correct*: mureo does not know that, so it reports both groups and names the campaign the borrowed scheme belongs to.

Four fixed, documented rules carry the false-positive story:

1. **Only `utm_*` is read at all** — a product id or variant flag in the URL is never compared on.
2. **Only `utm_source` / `utm_medium` / `utm_campaign` identify a campaign.** `utm_content` and `utm_term` exist so one campaign can tell its creatives and keywords apart on one landing page; comparing on them flags ordinary creative differentiation. They are still read (presence checks, declared patterns) but never make two ads disagree. `identify:` / `differentiate:` move a parameter between the sets.
3. **Schemes are compared as whole signatures, never one parameter at a time.** Per-parameter comparison reports "these ads borrowed campaign Y's `utm_source`" for a value like `google` that Y merely shares — and below three campaigns, or wherever one campaign carries a legitimate one-off ad, `google` *is* owned by exactly one other campaign, so the correctly-tagged majority gets flagged. A whole-signature match says something true instead.
4. **Digit runs are collapsed** when values are compared, so `segb01` ≡ `segb02` while `sega01` is different. This errs toward **fewer** findings.

A minimum campaign count was considered and rejected: the motivating incident happened in an account with exactly **two** Display campaigns, so a floor of three would switch the check off for the case it exists for. Rule 3 is what handles shared values instead.

## Severity reflects delivery state

A mis-tagged ad that has already served is a data-integrity incident needing a reporting caveat (`critical`); one that never served is a cheap fix (`high`). Delivery data that was not supplied is `delivery_state: unknown` with a note that the severity **may be understated** — never silently assumed. Omitting `impressions` is not the same as `0`.

## Enforcement — and exactly how far it reaches

| Path | Enforcement |
| --- | --- |
| `google_ads_ads_create`, `google_ads_ads_create_display` | **Enforced in the handler.** The check runs before the mutation; a finding refuses the create and no ad is uploaded. Reads `## Tracking Convention` from the workspace `STRATEGY.md` itself, so `identify:` / `differentiate:` apply here exactly as in the audit. Override per call with `acknowledge_tracking_findings=true`, or globally with `MUREO_DISABLE_TRACKING_PREFLIGHT=1`. **Fails open** on any error — and says so: the create comes back with `tracking_preflight: "NOT CHECKED: ..."`, every miss emits a log record, and three consecutive misses escalate to an ERROR log. Covers every way the check can fail to run, including the ones that raise nothing. One account-wide `list_ads` per create, cached per account for 60s so a bulk upload is a single read. |
| Meta Ads, plugin, bridged, hosted creates | **Routing only.** `_mureo-shared/SKILL.md` instructs the agent to run the check first; nothing stops it skipping the step. |
| `/tracking-health` | Account-wide audit, all platforms (step 8). |

**Not a `PolicyGate`, deliberately.** Its v1 ABI rules it out on its own terms: the Protocol is *synchronous by design* ("gates that need to await network I/O are out of scope for `mureo.policy_gates` in 0.9.x") and gates "MUST be pure and fast" since they run on every tool call — while this check needs one awaited platform read. A gate also never sees the campaign: `google_ads_ads_create` takes `ad_group_id`, not `campaign_id`, and no siblings, so a pure gate would have nothing to compare and would enforce nothing. The handler is the first point with both a client and an `await`.

The Meta gap is specific rather than an oversight: the destination link lives on the **creative**, created by a separate earlier call, so the ad-create arguments carry a `creative_id` and no URL. Tracked as follow-up rather than shipped half-done.

## Cross-platform by construction

The detector lives in core (`mureo/analysis/tracking/`) over a platform-neutral `AdTrackingRecord`, with one thin accessor per platform. It is exposed as the read-only `analysis_tracking_consistency_check` MCP tool, which **reaches no platform API** — the caller passes ad records in, so a platform mureo cannot fetch ads for is still auditable whenever the agent can list them.

| Platform | Where the URL lives | Read |
| --- | --- | --- |
| Google Ads | `final_urls` on `google_ads_ads_list` | yes |
| Meta Ads | creative `object_story_spec` link + creative `url_tags` (now requested by `meta_ads_ads_list`) | yes |
| Plugin platforms (Yahoo, LINE, SmartNews, LOGLY…) | provider ABI `Ad.final_url` | yes |
| Amazon Ads (bridged official MCP) | **not exposed** by the bridged tool surface | **no** — reported as unchecked |
| Hosted connectors | whatever their own list tool returns | best-effort, explicit field map |

**"No finding" is not "clean."** Ads whose URL could not be read are listed in `ads_without_readable_url` with a note, never reported clean. `records_from_mappings` requires the caller to state which field holds the URL rather than sniffing — a guessed field name that silently resolves to nothing would turn "not checked" into a clean bill of health.

## Tests

87 new tests. The **negative** ones are the point — a check that fires on `segb01` vs `segb02`, or on `utm_content=hero` vs `utm_content=video`, gets switched off by its first user:

- serial variation within one scheme (`segb01`..`segb08`); same LP, same scheme, different serial
- `utm_content` varying per creative on one landing page; `utm_term` varying per keyword; `utm_content` reused across two campaigns
- a two-campaign account where one campaign carries a legitimate one-off ad, and the three-campaign variant
- every campaign sharing `utm_source=google` + `utm_medium=cpc`
- two campaigns each internally consistent; two campaigns deliberately sharing one scheme
- per-landing-page word tokens, two ads each; untagged account; single-ad account
- a Google scheme and a Meta scheme are never compared
- and a regression test that the incident is **still** caught in a two-campaign account

Enforcement tests cover refusal, the acknowledgement override, the env kill switch, brand-new ad groups, three fail-open paths (including that the "NOT CHECKED" note is surfaced), the failure escalation and its reset-on-success, snapshot reuse across a bulk upload, and that a declared `differentiate:` / `identify:` changes the enforced verdict — including that the platform mutation is genuinely not awaited on refusal.

Platforms **covered by tests**: `google_ads` (detector + accessor + enforced create path), `meta_ads` (detector on Meta records and the creative accessor incl. `url_tags` merge and video CTA links), a provider-ABI plugin platform, and a generic bridged shape. Platforms **covered by design but untested**: every other plugin/bridge/hosted connector — they reach the same two accessors with different field names, and no live fixture exists here.

## Docs

`docs/tracking-consistency.md` documents what the check detects, what it does **not** — including two verified blind spots: campaign tokens differing only in digits (`campaign_2024` vs `campaign_2025` collapse to one shape, so sixteen mis-tagged ads give zero findings) and a source campaign carrying the scheme on a single ad when the campaigns use different landing pages — and, as explicitly, **what it may flag that you meant**. Also updated: `/tracking-health` + `_mureo-shared` + both platform foundation skills (in `skills/` and `mureo/_data/skills/`), `docs/mcp-server.md`, `docs/architecture.md`, `docs/strategy-context.md`, `AGENTS.md`, `README.md`, `README.ja.md`, CHANGELOG.

## Gates

`ruff check .` clean · `black --check .` clean · `mypy mureo --ignore-missing-imports` clean (299 files) · `pytest` in a **clean venv with no plugins**: **7879 passed, 8 skipped, 0 failed**.


hyoshi:
## Review fixes pushed (`400a2aa`)

All three CRITICALs and the HIGH are addressed. Clean-venv result up front, since that was the lesson: **7866 passed, 8 skipped, 0 failed** with no plugins installed — including `test_list_tools_returns_all_tools`. The 12 "known failures" I previously reported as an unchanged baseline were *entirely* environmental; none of them fail in a clean venv, so my baseline claim was worthless. Point taken: a test that already fails locally cannot tell you anything, and I will verify against a clean environment whenever a known-failure list overlaps what I touched.

### CRITICAL 1 — `utm_content` creative differentiation

Fixed by separating **campaign-identifying** from **creative-differentiating** parameters, as suggested. Only `utm_source` / `utm_medium` / `utm_campaign` take part in the scheme comparison. `utm_content` and `utm_term` are still read — presence checks and declared value patterns see them — they just never make two ads disagree. Both sets are adjustable: `identify:` promotes a parameter, `differentiate:` demotes one.

New negative tests: four creatives on one landing page varying `utm_content`; three ads varying `utm_term`; `utm_content` values reused across two campaigns; plus a positive test that `identify: utm_content` brings it back for an account that really does encode its segment there.

### CRITICAL 2 — the "exactly one other campaign" guarantee

The claim was false and I have removed it everywhere (`scheme.py`, `_checks_scheme.py`, `docs/`, `CHANGELOG`). Root cause was **per-parameter** comparison. Schemes are now compared as **whole identifying signatures**, so a value two campaigns merely share can never match an identity — `utm_source=google` only matters if `utm_medium` and `utm_campaign` match too, and if all three match, the ads genuinely are indistinguishable from campaign Y in reporting, which is the defect.

I did **not** add a minimum campaign count, and want to flag the disagreement explicitly rather than bury it: the motivating incident happened in an account with exactly **two** Display campaigns, so a floor of three would have switched the check off for the case it exists for. Your reasoning ("a shape shared by three campaigns is a house style") is right, but the signature comparison is what implements it without that cost. Tests pin all of it: the reviewer's two-campaign one-off repro, the three-campaign variant, shared source+medium across an account, and a regression test asserting the incident is *still* caught in a two-campaign account.

### CRITICAL 3 — tool count

205 → 206 in `tests/test_mcp_server.py`, `README.md:307`, `README.ja.md:273`, `docs/mcp-server.md:3` and `docs/architecture.md:459` (that last one I found while fixing the others).

### HIGH — enforcement

Now real for native Google Ads: `google_ads_ads_create` and `google_ads_ads_create_display` run the check in the handler before the mutation and **refuse** the create on a finding. Overridable per call with `acknowledge_tracking_findings=true` (same shape as `rollback_apply`'s `confirm`) or globally with `MUREO_DISABLE_TRACKING_PREFLIGHT=1`, and it **fails open** on any error reading the account — a tracking check that cannot see the account must not stop an operator shipping an ad. 14 tests, including that the platform mutation is genuinely not awaited on refusal.

**I did not wire PolicyGate, and I think that is the right call** — its own v1 ABI rules it out on two counts: the Protocol is *synchronous by design* ("gates that need to await network I/O are out of scope for `mureo.policy_gates` in 0.9.x") and gates "MUST be pure and fast" since they run on every tool call. This check needs one awaited platform read to compare the planned ad against the account. Worse, a gate never sees the campaign: `google_ads_ads_create` takes `ad_group_id`, not `campaign_id`, and no siblings — so a pure gate would have nothing to compare and would enforce nothing. The handler is the first point with both a client and an `await`. Happy to revisit if you want an async gate Protocol opened as separate work.

Meta, plugin, bridged and hosted creates stay **routing only**, now stated plainly in a per-platform enforcement table in the docs, the skill, and the CHANGELOG rather than implied to be covered. The Meta gap is specific: the destination link lives on the *creative*, created by an earlier call, so the ad-create arguments carry a `creative_id` and no URL.

### Also fixed

`docs/tracking-consistency.md` no longer claims an exhaustive limits list while showing only false negatives. It is now split into **What it does NOT detect** and **What it may flag that you meant**, the latter listing the deliberate-shared-identity case, digit-only-differing campaign tokens, and tracking A/B tests — plus a note that the two false positives you found are fixed and pinned rather than documented as limits.

hyoshi:
## Convention now reaches enforcement (`8243d0e`)

Clean venv: **7873 passed, 8 skipped, 0 failed**.

### HIGH — the enforcement path ignored `## Tracking Convention`

Confirmed and fixed. `google_ads_create_preflight` now loads `## Tracking Convention` from `STRATEGY.md` in the active workspace (mureo's own path — `<workspace>/STRATEGY.md`, never a caller-supplied string, so no traversal surface) and passes it to `preflight_tracking_consistency`. An absent, section-less or unreadable file falls back to the documented defaults, not to no enforcement.

Both directions are pinned by tests that **fail without the fix** — I verified that rather than assuming it:

```
FAILED ...::test_differentiate_stops_enforcement_blocking_it
FAILED ...::test_identify_extends_enforcement_to_a_custom_marker
2 failed, 3 passed        # with the convention= argument removed
5 passed                  # with it restored
```

There is also a control test asserting the same URL is still refused *without* the declaration, so the passing case cannot be passing for the wrong reason.

Your point about coverage landed. `_tracking_preflight.py` was at 100% the whole time and could not have been otherwise — the missing code was missing, and absent code has no uncovered lines. I've written that into the commit message so the next person reading the history gets it too. The generalisation I'm taking: coverage answers "did the tests execute this line", never "is the line that should be here here". The check that would have caught it is the one I skipped — comparing the advisory and enforcing paths on the same input and asserting they agree.

### Digit-collapsing blindness → moved to "does NOT detect"

You're right that I filed the consequence backwards. Reproduced before writing it down:

```
digit-only tokens, 8 mis-tagged ads          -> 0 findings
owner has 1 ad, different LPs, 16 mis-tagged -> 0 findings
```

`campaign_2024` and `campaign_2025` both collapse to `campaign_#`, so any account whose campaigns are distinguished by a year, month or sequence number is invisible to `foreign_campaign_scheme` — that is now an explicit false-negative entry stating the sixteen-ads-zero-findings result. The false-positive list keeps a shortened entry for the other direction of the same normalization, cross-referenced, since both come from one rule.

### `MIN_ADS_TO_OWN_A_SCHEME=2` + differing landing pages

Now its own explicit line in "does NOT detect", with the 16-vs-1 result, rather than a parenthetical in the table.

### Cost

Documented, and reduced. The read genuinely cannot be scoped to the target campaign — locating the campaign a scheme was copied *from* is what `foreign_campaign_scheme` does, so filtering to the campaign removes the check's reason to exist. Instead the account snapshot is cached per account for 60s (`SNAPSHOT_TTL_SECONDS`), which turns a bulk upload of 16 ads into **one** read — and a 16-ad burst is exactly the shape of the motivating incident. Tests assert 4 creates produce 1 `list_ads` call and that the cache is keyed per account. The staleness trade-off (an ad created inside the window is not visible to the next check) is documented.

### Fail-open observability

Agreed that `except Exception` alone makes a permanently broken guardrail indistinguishable from a quiet one. Two changes:

- A create whose check could not run now returns `tracking_preflight: "NOT CHECKED: <reason> …"` alongside the platform result — surfaced where the operator and agent actually look, not only in logs. The skill instructs the agent to report it and not treat it as a pass.
- Three consecutive failures escalate from `WARNING` to an `ERROR` reading "the create-time guardrail is effectively OFF; ads are being created unchecked". The counter resets on the first success.

The handler attaches the note to a **copy** of the client's result dict rather than mutating it.

### Gates

`ruff check .` clean · `black --check .` clean · `mypy mureo --ignore-missing-imports` clean (299 files) · clean venv, no plugins: **7873 passed, 8 skipped, 0 failed**.

hyoshi:
## MEDIUM fixed (`12d84c2`)

Clean venv: **7879 passed, 8 skipped, 0 failed**.

### The unresolved-campaign branch was invisible

You're right, and it undercut exactly the thing the previous commit advertised. `_record_failure` and `_not_run` are now a single `_fail_open(reason, *, exc=None)` that both branches go through: increment, log, return the note. So the branch that raises nothing now counts toward the same escalation as the ones that do — and that branch is the *most* important one to count, since an integration returning empty data instead of erroring is precisely the "broken guardrail looks like a quiet one" case.

The docstring on `_fail_open` says so explicitly, so the next person cannot re-split them without reading why they are joined.

Also restores the visibility regression you spotted: the branch logged at INFO before and had gone silent.

### Tests for the escalation and the reset

Added, and each verified to fail against the old behaviour rather than assumed to:

```
# unresolved branch reverted to a bare note:
FAILED ...::test_an_unresolvable_campaign_counts_towards_escalation
FAILED ...::test_every_not_run_emits_a_log_record

# reset-on-success deleted:
FAILED ...::test_a_success_resets_the_counter
```

`test_repeated_failures_escalate_to_error` asserts the exact `WARNING, WARNING, ERROR` sequence, which covers the `_FAILURES_BEFORE_ALARM` branch coverage reported as unreached. The STRATEGY.md read-exception branch is covered by a test that patches `Path.read_text` to raise and asserts the fallback lands on **defaults** — the declared `differentiate:` never loads, so the ad is still refused. That is the property worth pinning, not merely that the line executes.

One thing worth flagging from writing these: `caplog` collects from the root handler, so my first version counted a `mureo_agency.runtime_context` warning from this machine's environment as a pre-flight failure and the assertion was off by one. The tests now filter to `mureo.mcp._tracking_preflight`. In the clean venv it would have passed either way, which is the kind of test that passes in CI and lies locally.

### Test comment vs code

The code was right. The comment described `utm_medium` while the test declares `differentiate: utm_campaign`; rewritten to explain the test it is actually attached to — `utm_campaign` is the only parameter separating the two segments in that fixture, so differentiating it empties the identifying signature.

### `strategy_path`

Now taken from the active store's `strategy_path`. The `workspace / STRATEGY_FILENAME` form survives only as the fallback for a backend registered through `mureo.runtime_context_factory` that exposes no `strategy_path`, and a test pins that `_strategy_path()` follows the store when the store's path is moved — so the two cannot drift silently.

### Gates

`ruff check .` clean · `black --check .` clean · `mypy mureo --ignore-missing-imports` clean (299 files) · clean venv, no plugins: **7879 passed, 8 skipped, 0 failed**.

PR #572, opened 2026-08-07
feat: detect and diagnose delivery collapse across all platforms

### LINKED ISSUE #546: Detect and diagnose delivery collapse (the missing inverse of cost_increase_investigate), across all platforms
## Summary

mureo can detect a cost spike but not its inverse. `google_ads_cost_increase_investigate` exists; there is no counterpart for **delivery collapsing**. A campaign whose status is still ENABLED but whose impressions have gone to zero is the single most detectable failure mode in ad operations, and mureo currently has no scheduled detector and no root-cause workflow for it.

The generic `analysis_anomalies_check` is close but insufficient on its own: it must be hand-fed one campaign's current metrics, and its baseline comes from `action_log` history, which is empty or thin for accounts operated partly by hand.

Two Display campaigns went from roughly 350k impressions/day to zero while remaining ENABLED, and stayed at zero for about a week. Detection cost ~1.5 days. Diagnosis cost ~6 days, and never actually succeeded — the operator ended up rebuilding the campaigns rather than identifying the cause.

During the post-mortem, an agent with full API access ran campaign diagnosis, budget, ad approval, billing, bidding, and change-history checks and still could not fully close the question of why both campaigns died on the same day. If that is not answerable with complete API access, it is not answerable by an operator in the UI.

## Two separable pieces

**1. Detection (scheduled, unattended).** Flag any campaign whose delivery has fallen off a cliff relative to its own recent baseline while its status still says it should be serving. Impressions and spend, not conversions — `monitoring_zero_conversions` covers a different and much slower signal. The status-says-serving-but-nothing-is-happening contradiction is the whole point: it is what distinguishes a real fault from an intentional pause.

**2. Diagnosis (change x metric timeline).** Overlay the change feed on daily delivery metrics and answer "what changed immediately before the cliff?", then walk the standard elimination path: ad approval / policy, billing, budget, bid competitiveness, targeting and exclusion volume, learning state, campaign flight dates. Output should name the most likely cause with the evidence attached, not just list checks that passed.

The diagnosis half depends on the change-import work filed separately; without it the timeline has no change axis for manually-operated accounts.

## Cross-platform scope

This must not be a Google-only tool. The failure mode is universal:

- **Native platforms** (Google Ads, Meta Ads): implement directly, and register the capability on their existing `AnalyticsModule`.
- **Bridge / plugin platforms** (`plugin:<dist>` — LOGLY, LINE/Yahoo, SmartNews, Amazon official-MCP bridge): expose this as an advertised capability on the `mureo.analytics` entry-point group, so a provider can implement `diagnose_performance` / `detect_anomalies` for its own inventory model and have it run through `mureo_analytics_run`. Providers without a module report `analytics_not_available_for_<platform>` as today.
- **Hosted connectors** (`tiktok_ads`): the generic `analysis_anomalies_check` path already works for these when fed normalised metrics. The scheduled detector should feed it automatically rather than requiring an operator to do so by hand.

Suggested split so the platform-agnostic part is genuinely shared: cliff detection and the elimination ladder live in core and operate on normalised metrics; only the per-cause evidence lookups (policy status, billing state, learning state) are platform-specific.

hyoshi:
Closes #546

## What was missing

mureo could detect a cost spike but not its inverse. `google_ads_cost_increase_investigate` existed; nothing answered *"the campaign is still ENABLED and its impressions are zero."* That is the most detectable failure mode in ad operations, and it had no scheduled detector and no root-cause workflow.

The generic `analysis_anomalies_check` is close but insufficient: it must be hand-fed one campaign's current metrics, and its baseline comes from `action_log`, which is thin or empty on accounts operated partly by hand — silent exactly where the detector is needed.

## Two separable pieces

Detection and diagnosis ship as separate tools because they are worth having separately: knowing a campaign died is actionable hours before anyone knows why.

### 1. Detection (scheduled, unattended)

`analysis_delivery_collapse_check` (any platform, fed day-grain rows) plus a new `detect_delivery_collapse` analytics capability (Google Ads / Meta Ads, fetched automatically). `/daily-check` runs it in both modes.

**The baseline comes from the platform's own daily delivery** (`baseline_source: "platform_daily_delivery"`), never from `action_log`. `detect_delivery_collapse`'s signature is pinned by a test to `(series, thresholds, as_of)` so nobody reintroduces the empty-history blind spot.

**Not firing is as important as firing.** A detector that cries wolf gets muted, and a muted detector is worth nothing:

| Normal thing | How it is suppressed |
|---|---|
| Weekend / weekday seasonality | Baseline is the median of the **same weekday** in the window. A retail account at 350k on a Tuesday and 15k on a Saturday is a 96% swing — a flat median would raise a CRITICAL every weekend. |
| Intraday budget pacing | Days at or after `as_of` are partial by definition and are never evaluated. |
| Intentional pause | Status must be a serving status. The *status says serving, nothing is serving* contradiction **is** the signal. |
| Finished flight | A campaign past its `end_date` is expected to stop. |
| Low-volume campaign | Baselines under `delivery_collapse_min_baseline_impressions` (default 1000/day) hit zero routinely. |
| New campaign | Under `delivery_collapse_min_baseline_days` (default 14) of history → no signal. |

Thresholds are the operator's, in STRATEGY.md `## Guardrails` — nothing hard-coded at the call site:

```markdown
## Guardrails
- delivery_collapse_drop_pct: 90
- delivery_collapse_consecutive_days: 1
- delivery_collapse_min_baseline_impressions: 1000
- delivery_collapse_baseline_days: 28
- delivery_collapse_min_baseline_days: 14
- delivery_collapse_min_same_weekday_samples: 2
```

A bug found *by* the regression-injection pass and fixed here: the baseline window sits **strictly before the cliff**, so a collapse that outlives half the window cannot drag the median to zero and silence the alert. Judging each day against the window immediately preceding *it* would have gone quiet on exactly the week-long outage this was built for.

### 2. Diagnosis (change × metric timeline)

`analysis_delivery_collapse_diagnose` overlays the change feed on daily delivery — *what changed immediately before the cliff?* — then walks the elimination ladder: ad approval/policy → billing → budget → bid competitiveness → targeting and exclusions → learning state → campaign flight dates. `/rescue` gained the workflow that drives it.

**It does not promise a cause.** In the incident behind this issue, an agent with complete API access ran every one of those checks and still could not close the question. So the output is: what was **ruled out**, what is **implicated** with its evidence, and what remains **unresolved**. `most_likely_cause` is `null` and `confidence` is `undetermined` unless supplied evidence actually implicates a step — seven passing checks is an honest "undetermined", not a diagnosis.

**What it cannot answer** (returned verbatim on every response, in `limitations`):

- Serving-side suppression — the platform choosing not to enter a campaign into auctions — is not exposed by any read API mureo has, on any platform.
- No supported platform exposes billing state through an API mureo integrates.
- Learning-phase internals are not readable; a learning reset is inferred from a change event, never observed.
- Change feeds are incomplete: Google Ads change history omits system-initiated changes and retains ~30 days, Meta has no equivalent feed, and manual work reaches `action_log` only if imported. No change in the window is weak evidence, not exoneration.
- Several campaigns dying the same day is a correlation, never an attributed cause.

`next_checks` names the mureo tool for each open step where one exists, and an **empty string** where mureo has none (billing on every platform; bid competitiveness and learning state on Meta) rather than inventing one.

## Platform coverage — stated, not implied

| Platform | Detection | Diagnosis |
|---|---|---|
| **Google Ads** (native) | Automatic — `mureo_analytics_run` / `detect_delivery_collapse`; day-grain via new `get_daily_delivery_report` (live GAQL + BYOD) | Full ladder; change axis from `google_ads_change_history_list`; 5 of 7 steps have a mureo evidence tool |
| **Meta Ads** (native) | Automatic — same capability; day-grain insights joined to `effective_status` (live + BYOD) | Full ladder; **no change feed** (Meta has none — `action_log` only); 3 of 7 steps have a mureo evidence tool |
| **TikTok Ads** (hosted connector) | Core tool, fed normalised rows by the skill from TikTok's own reporting tools | Core ladder; no automatic evidence lookups |
| **Amazon Ads** (`plugin:mureo-amazon-ads-bridge`) | Core tool, fed rows from Amazon's own report tools. The bridge forwards Amazon's tools and declares no capabilities, so mureo cannot pull day-grain delivery itself | Core ladder; no automatic evidence, no change axis |
| **Yahoo / LINE / SmartNews / LOGLY** (plugin ABI) | Automatic **if** the provider implements the new `DeliveryCollapseModule`; otherwise the core tool fed rows | Core ladder; `next_checks` returns empty tools and an explicit unresolved note naming the platform |

The core detector, the baseline and the ladder are shared code; only the day-grain fetch and the per-cause evidence lookups are platform-specific.

## ABI note

`detect_delivery_collapse` is a **sibling** `DeliveryCollapseModule` Protocol, not a fifth method on `AnalyticsModule`. `AnalyticsModule` is `runtime_checkable`, and `isinstance` against a runtime-checkable Protocol requires every member — adding one would have silently made every already-published four-method plugin fail the check. The registry's structural validator is unchanged. Documented in `docs/ABI-stability.md` as the pattern for any future optional method.

`DeliveryCollapseReport.status` separates `ok` / `no_credentials` / `data_unavailable`, so a fetch failure is never read as an all-clear. The BYOD clients implement `get_daily_delivery_report` explicitly for the same reason: their `__getattr__` empty-list fallback would otherwise have passed a bundle off as "checked, nothing collapsed".

## Tests

TDD — tests written first and confirmed RED. 74 new tests. Ten regressions were injected into the implementation to prove the tests are not vacuous; **all ten were caught**, including the baseline-window bug above, which the injection surfaced as a real defect rather than a test gap.

## Gates

`ruff check .` clean · `black --check .` clean · `mypy mureo --ignore-missing-imports` clean · `pytest` — only the 12 environment-specific failures that already fail on `origin/main` locally (9 × `test_live_clients.py`, 3 × tool-registry counts inflated by a locally installed plugin).


hyoshi:
## Review fixes — both CRITICALs closed

### 1. Collapses past ~21 days went undetected

Reproduced exactly as reported. The measured sweep, before and after:

| | detected | missed |
|---|---|---|
| before | 1–21 | **22–180** |
| after | **1–180** | none |

`days_at_collapse` is exact across the whole range, and the same sweep on a weekday-seasonal account (350k weekdays / 15k weekends) is also clean, with the healthy control still silent.

Root cause was as diagnosed: the walk started at the most recent day, so on a long outage the *first* window it built was already mostly zeros, the baseline fell under `min_baseline_impressions`, and it broke on iteration one — never reaching the clean window further back. The scan now runs **forward** over every candidate and never breaks early, returning the earliest day from which delivery has stayed collapsed. Window baselines are memoised per weekday so it stays linear on healthy accounts.

On the test: you were right, and it is the part worth saying out loud. `test_a_collapse_longer_than_the_baseline_window_is_still_detected` used a **20**-day collapse against a **28**-day window — shorter than the window — while its name asserted the whole class was covered. It passed by a margin of a day or two and stopped anyone looking again. It now uses 45 days, and `test_collapse_duration_sweep` parametrises 1 → 180 including 21/22/23 so the boundary is asserted rather than accidentally cleared.

### 2. Sparse rows from the live APIs

Fixed defensively, without depending on the APIs' behaviour, as you asked. Every date from a campaign's first observed row through the **requested range end** with no row becomes an explicit zero. A platform that already returns zero rows leaves no gaps, so it is a no-op there — correct whichever way each API behaves, and it cannot rot if one changes.

Days *before* a campaign's first row are never invented: that would fabricate the very history the baseline is computed from (pinned by `test_fill_never_invents_days_before_a_campaign_first_appeared`).

Applied in all four clients, and again in `delivery_series_from_rows` — that second layer is what covers the hosted-connector / bridge / plugin path, where no mureo client exists to do it. Tests use mocked sparse responses for both Google and Meta and assert the detector then fires.

I did **not** verify the real API behaviour — no live credentials here — so this is the defensive reconciliation only, as instructed.

One consequence worth flagging rather than burying: if a platform has not yet materialised *yesterday's* row when the check runs, it is now filled as zero. The current (partial) day is never evaluated, so the exposure is bounded to a check run minutes after midnight; it is documented on the helper and in the skill, with `delivery_collapse_consecutive_days: 2` as the remedy.

### Also fixed

- **`min_baseline_days`** now counts days that actually **delivered**, not the window's raw length — 13 real days plus 3 already-collapsed ones is 13, not 16. Implementation and docstring now say the same thing.
- **`delivery_collapse_config.py`** — 53% → **100%**, including the `OSError` and unparseable-file branches. It exists solely to guarantee a broken STRATEGY.md cannot silently disable outage detection; that promise is now tested.
- **"Meta has no equivalent feed"** → "Meta publishes an account activity log but mureo does not fetch it yet." Corrected in `LIMITATIONS`, `docs/mcp-server.md` and the CHANGELOG. Fair catch — the feature's whole premise is that absence of evidence is not innocence, and that has to apply to our own gaps.
- **`byod/clients.py`** 1030 → **497** lines, split into `_client_common.py` (238) and `_meta_client.py` (396), with `clients.py` re-exporting both so every existing import path still resolves. One test's patch target moved with the function it patches (`_period_to_range`'s `date`).

### Exposed

`change_lookback_days` (default 3, 0–90) and `timeline_days` (default 21, 1–180) are now tool parameters. Your billing-hold example is the case: a cause with a delayed effect never reached `changes_before_cliff`. Changes outside the window still appear on the timeline — only the shortlist narrows.

### Verification

- Regression injection extended **10 → 19 mutations, all 19 caught**. The new ones cover the forward scan (`break` instead of `continue`; reversed iteration), the delivering-days gate, the gap-fill in all three places it lives, the config fail-open, and the diagnose lookback plumbing. One (`M14`) survived the first run — the second reconciliation layer was unpinned — and a test was added for it rather than the mutation being dropped.
- `ruff` / `black --check` / `mypy` clean. `pytest`: 7881 passed, only the same 12 environment-specific failures that already fail on `origin/main` here.

hyoshi:
## Lag CRITICAL closed

You were right on every count, including that my "bounded to minutes after midnight" claim was wrong. The exposure is reporting lag vs `consecutive_days`, not clock time — I reproduced it at 09:00 through `get_daily_delivery_report`, and at `consecutive_days=2` with a two-day lag.

### The matrix, before and after

**Healthy account — must be silent in every cell:**

| lag ↓ / `consecutive_days` → | 1 | 2 | 3 |
|---|---|---|---|
| 0 | silent → silent | silent → silent | silent → silent |
| 1 | **FIRES** → silent | silent → silent | silent → silent |
| 2 | **FIRES** → silent | **FIRES** → silent | silent → silent |
| 3 | **FIRES** → silent | **FIRES** → silent | **FIRES** → silent |

**Dead campaign (10 days) in a still-reporting account — must fire.** All 12 cells `days_at_collapse=10`, before and after. The fix does not give back the hole it was closing.

### What I chose, and why it differs slightly from your proposal

Your diagnosis of the root cause was exact: interior gaps are certain, trailing gaps are an assumption. But strict interior-only filling does **not** keep CRITICAL 2 closed, and this is the case you asked me to bring back if I found it: **a campaign that dies and stays dead produces only a trailing gap and nothing later.** That is the motivating incident itself — two Display campaigns to zero for a week. Interior-only would have missed exactly that.

So I moved the *bracket* rather than dropping it: the evidence is the **report**, not the single campaign. Any campaign in the account carrying a row for date D proves the platform covered D, so a campaign silent on D genuinely delivered nothing. Trailing days beyond the last date **anything** was reported are left out of evaluation entirely.

That keeps both criticals closed for the normal case (a dead campaign among live ones), at the cost of one residual: **when every campaign stops on the same day, nothing fires** — a total account outage and a platform-wide reporting failure are indistinguishable from there, and I am not willing to guess.

I did not hide that. `DeliveryCollapseReport` and `analysis_delivery_collapse_check` now carry `reported_through` and `unreported_days`; `/daily-check` treats 2+ (or a value that grew since yesterday) as **Action needed** — "the platform has reported no delivery since D". **An empty `signals` list is an all-clear only when `unreported_days` is 0**, and that sentence is now in the tool description, the docs and the skill.

Safe at defaults: no opt-in is required anywhere, and `delivery_collapse_consecutive_days: 2` is no longer a remedy for anything — it is back to being an ordinary sensitivity knob.

### Tests

`tests/test_delivery_collapse_reporting_lag.py` — 20 cases: single unreported day, two-day lag, the full 4×3 matrix on a healthy account, the same matrix with a dead campaign that must still be caught, interior-gap certainty, and the silent-account residual.

### Also

- **Mutation harness checked in** at `scripts/mutation_check.py` — the 20 mutations behind the claim, runnable and auditable (`--list`, substring filter, non-zero exit on a survivor). `tests/test_mutation_harness.py` asserts every anchor still resolves, and it immediately earned its place: it caught two anchors that `black` had reformatted out from under it. **20/20 caught**, including two new mutations for this fix (`fill_past_reported_range`, `scan_breaks_early`).
- **Blind spot documented**: a campaign with no rows anywhere in the window is invisible — no first row, no series, and mureo will not invent one. Now in `LIMITATIONS`, `docs/mcp-server.md` and `/daily-check`, not only in my head.
- **Bools rejected** on `change_lookback_days` / `timeline_days`, matching what the docs already promise for `analysis_anomalies_check`.
- **`_insights.py:214`** rewrapped.

Two tests from the previous commit asserted the now-wrong behaviour (filling to the requested end). Rewritten rather than deleted, so the correct rule is pinned and the single-campaign "report simply stops" case is asserted as no-signal plus a `reported_through` value.

### Gates

`ruff` / `black --check` / `mypy` clean. `pytest`: **7967 passed**, only the same 12 environment-specific failures that already fail on `origin/main` here.

Still no `python-reviewer` pass on any of the three commits — no `Agent` tool in my context. Given this round found a CRITICAL introduced by the previous round's fix, that gate is worth running before merge.

hyoshi:
## All three closed — `a36a19d`

### 1. The bracket's precondition, stated and tested

You were right that this was invisible: **every lag test I had used the same lag across all campaigns.** With mixed frontiers the faster campaign's latest date becomes the evidence for the slower one, and a healthy campaign gets zero-filled up to it:

```
a_lag0 reported through yesterday
b_lag2 reported through three days ago (alive)
-> b_lag2 flagged, days_at_collapse=2
```

The precondition — *every campaign in one call was fetched together and finalises at the same time* — is now written in four places a caller will actually hit: `fill_missing_delivery_days`' docstring, the `analysis_delivery_collapse_check` tool description, `docs/mcp-server.md`, and `/daily-check`. Two rules: pass all campaigns from **one** fetch, and when you cannot, set `reported_through` to the **oldest** per-campaign last date you trust.

**The remedy is reachable this time.** That was the lesson from the round before, so `reported_through` is now a parameter on `analysis_delivery_collapse_check`, not just a Python keyword argument. An agent following `/daily-check` can act on the guidance it is given.

Three tests: the hazard (pinned so it stays *known* rather than surprising), the remedy, and — importantly — that clamping the frontier costs a day or two of recency but **never hides a real collapse** (`dead` still detected at `days_at_collapse=12` alongside a clamped healthy `b_lag2`). Plus a mutation, `caller_frontier_ignored`, so the property is machine-checked.

I considered making this safe algorithmically rather than contractually — a quorum rule on the frontier — and rejected it: it is a new heuristic with its own failure modes (an account where half the campaigns die at once would step the frontier back and lose detection), invented under time pressure, changing semantics for every caller to fix a case the native connectors cannot produce. A stated precondition with a reachable escape hatch is the honest trade.

### 2. BYOD anchor drift

Restored the explicit anchor. It is computed from **every** metrics row before the campaign join, so inferring it afterwards let the frontier slip earlier whenever the rows carrying the bundle's last date belonged to a campaign missing from `campaigns.csv` — silently shortening every other campaign's window. A static bundle also cannot be behind on itself, so its own last date genuinely *is* the frontier: the one place asserting coverage is sound.

Regression test builds exactly that shape (c2 carries the newest dates and is absent from `campaigns.csv`; c1's 11-day outage must still be detected), parametrised across both BYOD clients, plus a `byod_anchor_inferred_after_join` mutation.

### 3. Test count — reconciled, and it was not a collection difference

**Both environments collect 7986.** Verified with `--collect-only` at `75e3c50`.

`tests/test_bridged_money_declarations.py::test_declared_paths_exist_in_the_manifest` skips when the local Amazon tool manifest file is absent. It exists on this machine (an agency setup wrote one) and does not in a clean venv, so that single test **passes here and skips there**:

| | passed | failed | skipped | collected |
|---|---|---|---|---|
| here | 7967 | 12 | 7 | 7986 |
| clean venv | 7978 | 0 | 8 | 7986 |

`7967 + 12 = 7979`, minus that one manifest-gated test = `7978`. Nothing hidden, no skipped module.

### Verification

- `pytest`: **7978 passed** here now (+11 new tests), same 12 environment-specific failures.
- Mutation harness: **22/22 caught**, up from 20.
- `ruff` / `black --check` / `mypy` clean.

Glad the harness proved it can say SURVIVED — that was the point of checking it in rather than asserting the number.

One thing still outstanding from my side, unchanged: no `python-reviewer` pass on any of the four commits, since I have no `Agent` tool in this context.

PR #573, opened 2026-08-07
Enforce learning-period reset awareness as a pre-flight check, not prose

### LINKED ISSUE #548: Enforce learning-period reset awareness as a pre-flight check, not prose (all auto-bidding platforms)
## Summary

`_mureo-shared/SKILL.md` documents "Learning Period Awareness" as a security rule — warn before changes that reset the learning period, list the affected operations, show current bidding system status. In practice this is **advisory prose addressed to the agent**, not an enforced pre-flight check. Nothing in mureo actually blocks or even reliably detects a learning-resetting change at the moment it is dispatched.

A campaign whose delivery had already collapsed was "fixed" by changing its bid ceiling. That change put the bidding strategy into a settings-change learning state, which delayed recovery further. The operator was working blind, trying things — which is exactly the state in which this guard matters most, and exactly the state in which an agent is least likely to remember a prose rule.

The general pattern: learning resets are most damaging precisely when someone is troubleshooting a broken campaign, because troubleshooting means making many changes quickly.

## Proposed shape

A pre-flight check on dispatch that (a) reads the campaign's current learning state, (b) determines whether the pending change is in the reset-triggering class, and (c) surfaces both in the confirmation step, with the option to express a hard refusal in `STRATEGY.md` `## Guardrails` (e.g. `block_learning_resets_during_incident`).

Escalation is warranted when the campaign is *already* in a learning state — stacking a second reset on an unstable campaign is almost never intended, and should require an explicit override rather than a default-yes confirmation.

## Cross-platform scope

Learning phases are not a Google concept; every major automated-bidding system has one, and the reset triggers differ per platform, which is exactly why this needs per-platform rules behind one core interface:

| Platform | Learning concept | Typical reset triggers |
|---|---|---|
| Google Ads | `bidding_strategy_system_status` (e.g. learning after settings change) | bid strategy / target changes, large budget moves, conversion setting changes |
| Meta Ads | ad set learning phase, "Learning Limited" | significant edits to budget, bid, targeting, creative; ad set restarts |
| Yahoo Ads | auto-bidding adjustment period | bid strategy and target changes |
| LINE / SmartNews Ads | auto-bid adjustment | strategy / target changes |
| LOGLY (`plugin:mureo-logly-bridge`) | tCPA readiness — `logly_ads_context_tcpa_readiness` already exposes a readiness concept | tCPA target changes, bid merges |
| Amazon Ads (official-MCP bridge) | bid optimisation adjustment | strategy changes |

Suggested design: core owns "is this change class reset-triggering, and what is the current learning state" as an interface; each native platform implements it; bridges and plugins can advertise it as an optional capability (a natural fit alongside the existing `mureo.analytics` capability set, or as a provider ABI method). Platforms that cannot report learning state should say so rather than silently returning "safe" — a false negative here is worse than no answer, because it converts a warning into implied approval.

hyoshi:
Closes #548

## What was wrong

`_mureo-shared/SKILL.md` carried "Learning Period Awareness" as a **security rule**: warn before changes that reset the learning period, list the affected operations, show the current bidding system status. Nothing in mureo detected — let alone blocked — a learning-resetting change at the moment it was dispatched.

A prose rule is followed worst exactly when it matters most. The incident behind #548: a campaign whose delivery had already collapsed was "fixed" by moving its bid ceiling, which put the bidding strategy into a settings-change learning state and delayed recovery further. Troubleshooting *means* making many changes quickly.

## Three surfaces, of deliberately different strength

MCP has no interposed confirmation step — mureo either runs a tool call or refuses it — so these are not interchangeable, and the PR does not pretend they are.

| Surface | When | Strength |
|---|---|---|
| `## Guardrails` `block_learning_resets` / `block_learning_resets_during_incident` | before dispatch | **hard** — refused by the existing `StrategyPolicyGate`, before any API call |
| `mureo_learning_reset_preflight` (new tool, read-only) | before the change, when the agent calls it | advisory — as strong as the agent's compliance |
| Notice appended to a reset-triggering call's own result | after that call | records the reset so the *next* change in the sequence is not made blind |

It runs on the **one existing gate**, not a parallel mechanism: `evaluate_guardrails` gained an optional `learning=` argument and stays I/O-free; the gate does the STATE.json read, and only when the operator actually wrote one of the two rules. Both default off, so the fail-open contract is intact — with no rule written, behaviour is byte-identical and no STATE.json read happens.

`would_block` in the pre-flight tool is computed by the same `learning_reset_denial` the gate calls, so the report and the enforcement cannot drift.

## Evidence, not folklore

Reset triggers are **not hardcoded from recollection**. `mureo/policy/learning_rules.py` carries, per trigger: the first-party source URL, the retrieval date, and the sentence it rests on.

Google Ads is complete because Google publishes the causes as an **enum**: every `LEARNING_*` member of [`BiddingStrategySystemStatus`](https://developers.google.com/google-ads/api/reference/rpc/v23/BiddingStrategySystemStatusEnum.BiddingStrategySystemStatus) names the change class that caused re-entry — and that exact enum ships inside the `google-ads` client this repo already depends on, so the quotes are checkable offline. The same enum is the state read, including `UNAVAILABLE` → *"does not support status reporting"* → reported `unreportable`, not `steady`.

The old prose's `budget changes > 20%` threshold has **no** first-party source. It is gone rather than reproduced; Google's enum says "a recent budget change" with no magnitude, so mureo applies none. That, and the other honest limits, are written down in the rules table's `notes` rather than smoothed over.

## Per-platform coverage — what is real

| Platform | Learning state readable by mureo? | Reset triggers known? |
|---|---|---|
| **Google Ads** | **Yes** — `bidding_details.bidding_strategy_system_status` on the campaign's STATE.json snapshot | **Yes** — Google's `LEARNING_*` enum members |
| **Meta Ads** | **No** — Meta exposes [`learning_stage_info`](https://developers.facebook.com/docs/marketing-api/reference/ad-campaign-learning-stage-info/) on the **ad set**; mureo's client does not request it and STATE.json is campaign-level | **No** — Meta documents that "significant edits" restart the phase, and exposes `last_sig_edit_ts`, but does not enumerate which edits are significant |
| **Amazon Ads** (`plugin:mureo-amazon-ads-bridge`) | **No** — no learning-state read exists anywhere on the bridged tool surface mureo knows (checked against `amazon_ads/reversal.py`, `amazon_ads/money_paths.py`, `docs/amazon-ads.md`) | **No** |
| **Yahoo / LINE / SmartNews / LOGLY** (plugins) | **No** — unless the plugin registers rules | **No** — unless the plugin registers rules |

So the check is **complete on Google Ads and honest everywhere else**. A platform mureo cannot classify reports `unknown`; a platform whose state it cannot read reports `unreportable`. Neither is ever `no_reset` / `steady` — a false "this resets nothing" converts a missing warning into implied approval, which is worse than no answer.

Bridges and plugins advertise their own rules through `register_platform_learning_rules`, the same registry pattern the budget/bid declarations already use.

## Not a false-alarm generator

A check that fires on everything gets ignored, so the classifier is argument-aware and read-aware:

- `google_ads_campaigns_update` with `name` → `no_reset`; the **same tool** with `bidding_strategy` → `resets`.
- `google_ads_campaigns_update_status` with `PAUSED` → `no_reset`; with `ENABLED` (reactivation) → `resets`.
- Every read on every platform → `no_reset` (needs no per-platform evidence).
- The appended notice fires **only** on `resets`, never on `unknown` — otherwise it would fire on every Meta and Amazon mutation. The pre-flight tool still reports `unknown` honestly when asked.

## Where the state comes from, and why

A `PolicyGate` runs on every tool call and must be pure and fast, so the pre-flight **cannot** ask the platform. It reads what mureo already has locally: the STATE.json campaign snapshot (`google_ads_campaigns_get` / `_diagnose` → `mureo_state_upsert_campaign`). A stale or missing observation is reported `unknown`, and `block_learning_resets_during_incident` treats `unknown` as "refuse" rather than "fine".

## Tests

42 new tests in `tests/test_learning_reset_preflight.py`, organised around the four properties that matter: the reset verdict and learning state are surfaced at dispatch; a `## Guardrails` refusal actually **blocks** (end-to-end through `StrategyPolicyGate` and through `handle_call_tool`); a non-resetting change is **not** warned about; an unreadable platform says so.

Not vacuous — five regressions were injected and each was caught:

| Injected bug | Tests that failed |
|---|---|
| unknown learning state counted as "not learning" | 2 |
| guardrail reports but never blocks | 6 |
| notice fires on every mutation | 1 |
| unclassifiable platform reported `no_reset` | 3 |
| trigger matching ignores arguments (rename warned) | 3 |

## Docs

Updated with the code: `_mureo-shared/SKILL.md` §6 rewritten from prose rule into a description of the mechanism (including the per-platform coverage table and the explicit statement that `unknown` never means safe), `_mureo-strategy/SKILL.md` guardrail reference, `docs/mcp-server.md` (new section + tool count 205 → 206), `docs/architecture.md`, both READMEs, `AGENTS.md`, `CHANGELOG.md`, and the `bidding_details` schema description in `mureo_state_upsert_campaign` so the convention is discoverable from the tool itself. Both skill copies (`skills/` and `mureo/_data/skills/`) stay in sync.

## Gates

`ruff check .` clean · `black --check .` clean · `mypy mureo --ignore-missing-imports` clean · `pytest` 7815 passed, 12 pre-existing local-environment failures only (9 × `test_live_clients.py`, 3 × tool-registry counts caused by extra plugins installed on the dev machine), no new failures.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,7 @@ mureo/
 │   ├── tools_analytics_registry.py        # mureo_analytics_modules_list / mureo_analytics_run (#440)
 │   ├── tools_creative_studio.py           # creative_studio_* (visual generation + compose)
 │   ├── tools_learning.py                  # mureo_learning_insights_get / mureo_consult_advisor
+│   ├── tools_learning_preflight.py        # mureo_learning_reset_preflight (#548)
 │   ├── tools_mureo_context.py             # STRATEGY.md / STATE.json + mureo_outcome_evaluate tools
 │   ├── _handlers_mureo_context.py         # Context (STRATEGY/STATE) handlers
 │   ├── _client_factory.py                 # Per-platform BYOD-vs-live client factory
@@ -133,6 +134,8 @@ mureo/
 ├── core/                # Extension Protocols + file-backed impls + RuntimeContext; provider & skill discovery
 ├── providers/           # Official MCP provider catalog + one-command install helpers (#86)
 ├── policy/              # Built-in policy gates (strategy_gate) — ship with OSS, run by default
+│   ├── learning_rules.py     # Per-platform learning-period facts + their first-party sources (#548)
+│   └── learning_reset.py     # Reset-class + learning-state pre-flight the gate refuses on
 ├── learning/            # Read-side /learn companion: insight federation across configured sources
 ├── creative_studio/     # Creator-grade ad-creative (image) generation via pluggable providers
 ├── byod/                # Bring Your Own Data — CSV-backed offline analysis (see BYOD Mode below)
@@ -227,6 +230,7 @@ These families are not tied to a single ad platform. Tool names are the exact MC
 | Analysis | `analysis_anomalies_check` |
 | Creative Studio | `creative_studio_providers_list`, `creative_studio_generate_visual`, `creative_studio_edit_visual`, `creative_studio_compose`, `creative_studio_brand_kit_get` |
 | Learning | `mureo_learning_insights_get`, `mureo_consult_advisor` |
+| Learning pre-flight (#548) | `mureo_learning_reset_preflight` |
 | mureo Context | `mureo_strategy_get`, `mureo_strategy_set`, `mureo_state_get`, `mureo_state_action_log_append`, `mureo_state_upsert_campaign`, `mureo_state_report_set`, `mureo_state_platform_metrics_set`, `mureo_state_set_conversion_events`, `mureo_outcome_evaluate` |
 
 ## Design Constraints
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,10 +97,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - **Hard refusal.** Two new STRATEGY.md `## Guardrails` rules, enforced by the
     existing built-in `StrategyPolicyGate` before dispatch (no second gate):
     `block_learning_resets` refuses every reset-triggering change;
-    `block_learning_resets_during_incident` refuses one unless the campaign is
-    *positively known* to be out of a learning period — so it also fires on an
-    unknown state, matching the gate's existing discipline of refusing what it
-    cannot verify. Both default off: with neither written, behaviour is
+    `block_learning_resets_during_incident` is narrower by name and in fact —
+    it refuses only a change that *identifies a campaign* which is not
+    positively known to be out of a learning period. An unknown state on an
+    identified campaign is refused (fail-closed, matching the gate's existing
+    discipline of refusing what it cannot verify); a change that identifies no
+    campaign at all has no subject and is not refused, since several
+    reset-triggering tools are account-level (`google_ads_conversions_*`) or
+    keyed on something other than a campaign (`google_ads_budget_update` takes
+    a `budget_id`) and would otherwise be blocked forever with no relation to
+    any incident. Both default off: with neither written, behaviour is
     unchanged and mureo neither reads STATE.json nor blocks anything.
   - **Pre-flight tool.** `mureo_learning_reset_preflight` (read-only, no
     platform API call) answers, *before* the change: is it reset-triggering,
@@ -131,9 +137,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Reads and unrelated mutations are never flagged: a campaign rename through
   `google_ads_campaigns_update` is `no_reset` while the same tool carrying
-  `bidding_strategy` is `resets`, and pausing a campaign is `no_reset` while
-  re-enabling it is `resets`. A check that fires on everything is a check that
-  gets ignored.
+  `bidding_strategy` is `resets`; renaming a conversion action through
+  `google_ads_conversions_update` is `no_reset` while changing its category,
+  status, value or lookback window is `resets`; and pausing a campaign is
+  `no_reset` while re-enabling it is `resets`. A check that fires on everything
+  is a check that gets ignored — and one that blocks a rename is a check that
+  gets switched off.
+
+  The classification's known blind spots (a manual-CPC campaign has no bid
+  strategy to reset, so `budget_change` / `composition_change` / `reactivation`
+  are over-reported on one) travel with the **`resets`** verdict, not only with
+  `no_reset` — the operator stopped by a false positive is the one who needs
+  them.
 
 ## [0.10.43] - 2026-08-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,62 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   construction at the tool. Every other removal stays blocked, and both
   invariants are pinned by tests.
 
+- **Learning-period reset pre-flight — an enforced check, not prose** (#548).
+  `_mureo-shared/SKILL.md` documented "Learning Period Awareness" as a security
+  rule: warn before changes that reset the learning period, list the affected
+  operations, show the current bidding system status. Nothing in mureo actually
+  detected or blocked such a change at the moment it was dispatched — it was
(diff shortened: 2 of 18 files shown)


hyoshi:
## Review fix — HIGH addressed (+ the caveat-placement point)

### 1. Pure rename no longer classified as a reset

`google_ads_conversions_update` carried no `requires_arguments`. Google publishes **two** conversion causes, so the trigger is now split to match the evidence rather than patched:

| change_class | tools | condition | enum quoted |
|---|---|---|---|
| `conversion_type_change` | `conversions_create`, `conversions_remove` | unconditional (neither has a cosmetic-only mode — narrowing must not become under-reporting) | `LEARNING_CONVERSION_TYPE_CHANGE` |
| `conversion_settings_change` | `conversions_update` | `category` / `status` / `default_value` / `always_use_default_value` / `click_through_lookback_window_days` / `view_through_lookback_window_days` — every updatable field **except `name`** | `LEARNING_CONVERSION_SETTING_CHANGE` |

`status` is in the settings set deliberately: HIDDEN / REMOVED / ENABLED decides whether the action counts toward "Conversions", which is the reporting the bid strategy depends on.

### 2. `block_learning_resets_during_incident` now requires a subject

"During incident" names a *specific campaign known to be unstable*. It refuses only when the call **identifies a campaign** and that campaign is not positively known to be out of a learning period.

- Fail-closed **kept where it has a subject**: `unknown` on an *identified* campaign is still refused.
- Fail-closed **kept whole** for `block_learning_resets` — that one is honestly an account-wide freeze and needs no subject.

Without this, every account-level tool (`google_ads_conversions_*`) and every non-campaign-keyed one (`google_ads_budget_update` takes a `budget_id`) resolved to no campaign, state always `unknown`, refused forever. Reached by *following* the notice text that recommends the rule.

### The rename case, before and after

```
build_preflight("google_ads_conversions_update",
                {"conversion_action_id": "CA1", "name": "Renamed"}, StateDocument())
```

| | risk | change_class | campaign_id | `during_incident` blocks | `block_all` blocks |
|---|---|---|---|---|---|
| **before** | `resets` | conversion_settings_change | None | **yes** | yes |
| **after** | `no_reset` | — | None | **no** | **no** |

Neighbouring cases confirmed unchanged in the right direction:

| call | risk | `during_incident` | `block_all` |
|---|---|---|---|
| `conversions_update` `status=REMOVED` (archive) | `resets` | no — no campaign subject | **yes** |
| `conversions_update` `default_value=5000` | `resets` | no — no campaign subject | **yes** |
| `campaigns_update` `bidding_strategy`, state `unknown`, campaign `C1` | `resets` | **yes** — subject identified, fail-closed | **yes** |

### 3. Caveats moved to where the refusal is

The blind-spot note was only ever emitted on the `NO_RESET` path — the operator *blocked by a false positive* never saw it. It now travels with the `resets` verdict, so it reaches the refusal reason and the dispatch notice. Widened as requested: it now names **`budget_change`, `composition_change` AND `reactivation`** as sharing the manual-CPC blind spot, not just `budget_change`.

Verified in the refusal text: `"manual-CPC" in reason == True`.

### Tests

54 total (12 new). Each new behaviour proved non-vacuous by injection:

| Injected bug | Test that failed |
|---|---|
| revert the campaign-subject guard | `test_incident_guardrail_needs_a_campaign_subject` |
| settings trigger matches `name` again | `test_conversion_rename_is_not_a_reset` |
| caveat back on `NO_RESET` path only | `test_reset_verdict_carries_the_classification_caveats` |
| note narrowed back to `budget_change` | `test_reset_verdict_carries_the_classification_caveats` |

Also pinned: all six settings fields individually → `resets`; `create`/`remove` stay `resets`; `block_all` still refuses a subject-less change.

### Gates

`ruff` clean · `black --check` clean (610 files) · `mypy` clean (291 files) · `pytest` **7827 passed**, 12 pre-existing local-environment failures only, no new failures.

Docs updated with the semantics change: `_mureo-strategy/SKILL.md` (both copies), `docs/mcp-server.md`, `CHANGELOG.md`.

PR #574, opened 2026-08-07
Size a bulk exclusion against the account's own delivery before applying it

### LINKED ISSUE #547: Show delivery impact before applying bulk exclusions / blocks, and let it be a guardrail
## Summary

mureo will happily apply a bulk exclusion / block / negative-targeting change without telling the operator how much of their delivery it is about to remove. There is no "you are about to exclude inventory that accounted for 94% of your last 30 days of impressions" step.

This is the only requested feature that prevents the originating mistake rather than shortening the recovery from it.

An operator tightened Display placement exclusions incrementally over two weeks to improve engagement quality, then did one larger pass. Delivery went to zero and stayed there for about a week. The intent was reasonable and the direction was correct; the magnitude was not visible at the moment of the decision.

## Proposed shape

Before applying a batch of exclusions / blocks / negative keywords, compute the share of the recent window's impressions, clicks, spend and conversions attributable to the entities being excluded, and surface it in the confirmation step that already exists for write operations. Warn above a configurable threshold, and make the threshold expressible in `## Guardrails` in `STRATEGY.md` so it can be enforced rather than merely displayed.

The estimate should be based on the account's own recent performance data, not a platform reach estimator — the question is "how much of *my current delivery* does this remove", which is directly computable.

Reporting "coverage unknown" is an acceptable and honest output where the platform cannot attribute past delivery to the entities being excluded. Silently proceeding is not.

## Cross-platform scope

Every platform mureo touches has an inventory-restriction surface, so this belongs in core with per-platform data sources:

| Platform | Exclusion surface | Performance source for the estimate |
|---|---|---|
| Google Ads | excluded placements / mobile apps / app categories, negative keywords | placement & search-term performance reports |
| Meta Ads | publisher platform / placement restriction, block lists | `meta_ads_analysis_placements`, insights breakdowns |
| Yahoo Ads (Display) | placement URL lists (`yahoo_ads_display_*_placement_url_list`) | display daily report by placement |
| LINE / SmartNews Ads | placement & category restrictions | their insights / report tools |
| LOGLY (`plugin:mureo-logly-bridge`) | adspot blocking, adspot bid merges | `logly_ads_context_adspot_daily_report` **plus** the supply-side view |
| Amazon Ads (official-MCP bridge) | negative targeting | reporting tools |

**The LOGLY case is worth calling out specifically**, because the account knowledge base already contains a hard-won operator insight that says exactly this: never block an adspot on your own campaign's numbers alone — check `logly_supply_adspot_summary` for the slot's all-advertiser CTR/CPC, and `logly_supply_adspot_cv_by_category` for whether other advertisers convert there, because your own data cannot distinguish "bad inventory" from "my creative mismatches this inventory".

That insight is currently enforced only by an agent reading prose and choosing to comply. This issue is the mechanism that makes it structural: the impact preview is the natural place to require the supply-side check before a block is applied. Providers should be able to contribute their own enrichment to the preview through the provider ABI so the same discipline applies to any bridge platform that has a supply-side view.

## Note on Google Ads specifically

Google Ads placement/app/app-category exclusions cannot currently be executed through mureo at all (filed separately). That issue is a prerequisite for this one on Google — but this issue should not wait for it, since Yahoo, LOGLY and negative keywords are already actionable surfaces today.

hyoshi:
Closes #547

Based on `feat/placement-exclusion-tools` (#544) — this guards the exclusion tools that branch adds.

## What was wrong

mureo would apply an exclusion / block / negative-keyword batch without telling the operator that the inventory it was about to remove carried 94% of the last 30 days of impressions. Of everything filed from the incident post-mortem, **this is the only item that prevents the originating mistake rather than shortening the recovery from it**: the operator tightened Display placement exclusions over two weeks, then made one larger pass, and delivery went to zero for a week. The intent was reasonable and the direction was correct — the magnitude was never visible at the moment of the decision.

## The correction to the issue's premise

The issue asks for this "in the confirmation step that already exists for write operations". **There is no such step.** MCP gives mureo two outcomes — run the call or refuse it — and `PolicyDecision` has no allow-with-notice path (the `reason` is discarded when a gate allows). So the ask becomes three surfaces of deliberately different strength, and the PR does not pretend they are interchangeable:

| Surface | When | Strength |
|---|---|---|
| `## Guardrails` `max_delivery_share_removed_pct` / `max_cumulative_delivery_share_removed_pct` / `block_exclusions_without_impact_data` | before dispatch | **hard** — refused before any API request reaches the platform |
| `analysis_exclusion_impact_preview` (new tool, read-only) | whenever the agent asks | advisory — as strong as the agent's compliance |
| Notice appended to an allowed exclusion's own result | after that call | records the measured size, so the **next** pass in an incremental sequence is not made blind |

The third one is what the motivating incident actually needed. The damage was two weeks of individually-small passes and not one of them was ever measured.

## Where it is enforced, and why not in the gate

**In the dispatcher (`mureo/mcp/exclusion_preflight.py`), not in `StrategyPolicyGate`.** Sizing an exclusion needs one *awaited* read of the account's own recent delivery, and the `PolicyGate` v1 ABI is *synchronous by design* ("gates that need to await network I/O are out of scope for `mureo.policy_gates` in 0.9.x") and "MUST be pure and fast" because it runs on every tool call. So the keys are parsed by the gate's **own parser** — one `## Guardrails` vocabulary, one parser — and enforced at the first point in dispatch that has both an `await` and the arguments, before `_dispatch_tool`, so a refusal lands before any mutation.

`Guardrails.is_empty()` deliberately stays blind to the new fields: counting them would switch the gate's budget/bid pattern scan (and its fail-closed deny on an exhausted scan) on for an operator who wrote only an exclusion rule and no money cap at all.

**Fail-open and cheap when off.** With none of the three rules written the pre-flight returns before it builds a client: no report request is issued, behaviour is byte-identical. A tool that is not a registered exclusion surface returns even earlier, without touching STRATEGY.md. `MUREO_DISABLE_EXCLUSION_PREFLIGHT=1` turns it off entirely.

## Two figures, because one is not enough

`incremental` is what this batch removes. `cumulative` is what the account's **whole** standing exclusion set removes once this batch lands.

That is what catches the incremental tightening: an entity excluded a week ago still carries its pre-exclusion impressions inside a 30-day window, so a fortnight of small passes shows up cumulatively even when no single pass trips the incremental cap.

Its honest limit: **an exclusion older than the window contributed nothing to it and is invisible**, so the cumulative share is a lower bound. And where mureo cannot list the standing set it is **withheld** (`null` with a reason), never reported as zero — that happens for an ad-group-level Google placement write (campaign-level exclusions also cover that ad group and are not reachable from the call's arguments) and for `google_ads_negative_keywords_add_to_ad_group` (Google Ads exposes no ad-group-level negative keyword listing).

## Per-platform, and honest where it cannot answer

| Surface | Delivery source | Attributable? |
|---|---|---|
| `google_ads_negative_placements_add` | new `group_placement_view` read (`get_placement_performance`) | **Yes** for `website` / `mobile_application`. A `mobile_app_category` in the same batch is **not** — a category is not a placement that serves — so a mixed batch reports `partial`, not a confident number |
| `google_ads_negative_keywords_add` / `_add_to_ad_group` | `search_term_view` over the same window | **Yes**, matched per `EXACT` / `PHRASE` / `BROAD`. Negative keywords do not match close variants and neither does the estimate, so it is a lower bound |
| `meta_ads_excluded_placements_set` | — | **No.** No Meta insights breakdown attributes past delivery to publisher categories, publisher block lists or brand-safety content types. `publisher_platform` / `platform_position` *are* attributable but are a different exclusion surface (ad-set targeting). Reported `unknown` |
| Yahoo / LINE / SmartNews / LOGLY (plugins), Amazon (bridge) | whatever the provider registers via `register_exclusion_surface` | **Provider-declared.** mureo registers nothing on their behalf rather than guessing a bridged tool's argument shape and producing a confident wrong number |

`unknown` is an acceptable output; a silent "no impact" is not. Coverage is `measured` / `partial` / `unknown`; `incremental` is `null` rather than a row of zeroes when nothing can be attributed; a window that served nothing reports `share_pct: null` rather than `0`, because a printed `0%` reads as approval. Even when nothing blocks, the measured — or unmeasurable — verdict is appended to the call's own result, so an exclusion is never applied *silently*.

`analysis_exclusion_impact_preview` also takes `excluded_entities` + `delivery_records` directly, in which case it **reaches no platform API at all** — so a Yahoo placement URL list, a LOGLY adspot block or an Amazon negative-targeting batch is auditable whenever the agent can pull that platform's own report. `would_block` is computed by the same `evaluate_exclusion_impact` the pre-flight calls, so the advertised verdict and the enforced one cannot drift.

The LOGLY discipline the issue calls out (never block an adspot on your own campaign's numbers alone — check `logly_supply_adspot_summary` and `logly_supply_adspot_cv_by_category`) is written into `_mureo-shared/SKILL.md` next to the preview step: the impact preview sizes the loss, the supply-side view says whether the loss is worth taking. The `ExclusionSurface` registry is the ABI hook for a bridge that wants to contribute its own supply-side enrichment.

## Not a false-alarm generator

A check that fires on everything gets switched off, so every matching rule errs toward **fewer** attributed impressions:

1. Excluding `example.com` covers `news.example.com` (Google excludes the site), but excluding `news.example.com` never claims `example.com`, and the label-boundary check stops `example.com` swallowing `notexample.com`.
2. Negative keywords are matched by token sequence per match type, with **no** stemming, pluralisation or accent folding — Google does not apply close variants to negatives either.
3. Any other entity kind is exact-string, case-folded.
4. The cap applies only to the metrics the operator listed (`exclusion_impact_metrics`, default `impressions`).

## New `## Guardrails` keys

All optional, all default off:

```markdown
## Guardrails
- max_delivery_share_removed_pct: 25
- max_cumulative_delivery_share_removed_pct: 60
- exclusion_impact_window_days: 30
- exclusion_impact_metrics: impressions, cost
- block_exclusions_without_impact_data: false
```

## Tests

38 new in `tests/test_exclusion_impact_preview.py`, organised around the four properties that decide whether this works: the share is computed and surfaced; a `## Guardrails` threshold **actually blocks**, end to end through `handle_call_tool` and before the mutation; a small exclusion is neither warned about nor blocked; an unattributable surface says `unknown` rather than passing silently. Plus 4 in `tests/test_google_ads_negative_placements.py` for the new `group_placement_view` read (including GAQL-validator coverage of the scope id and the window).

Not vacuous — seven regressions were injected and each was caught:

| Injected bug | Tests that failed |
|---|---|
| share cap reports but never blocks | 5 |
| unknown / zero-total coverage rendered as `0%` | 1 |
| host match drops the label boundary | 1 |
| pre-flight reads the platform with no rule written | 1 |
| `block_exclusions_without_impact_data` ignored | 2 |
| missing standing list read as an empty one | 1 |
| unattributable entity kinds silently dropped | 1 |

Full suite green in a **clean venv** (7883 passed, 8 skipped) — this machine has plugins installed, so the local tool-count result is not evidence. `ruff check .`, `black --check .`, `mypy mureo --ignore-missing-imports` all clean.

## Tool count

210 → 211 (Analysis 1 → 2). README (both), `docs/mcp-server.md`, `docs/architecture.md`, `AGENTS.md`, the four foundation skills and their `_data` mirrors updated.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,8 +92,11 @@ mureo/
 │   ├── _handlers_search_console.py        # Search Console handlers
 │   ├── tools_rollback.py                  # rollback_plan_get / rollback_apply
 │   ├── _handlers_rollback.py              # Rollback handlers (lazy-resolve dispatcher)
-│   ├── tools_analysis.py                  # analysis_anomalies_check
+│   ├── tools_analysis.py                  # analysis_anomalies_check / analysis_exclusion_impact_preview
 │   ├── _handlers_analysis.py              # Anomaly detector composition handler
+│   ├── _handlers_exclusion_impact.py      # analysis_exclusion_impact_preview handler (#547)
+│   ├── exclusion_preflight.py             # Pre-dispatch exclusion sizing + refusal + notice (#547)
+│   ├── exclusion_sources.py               # Built-in per-platform delivery sources for the above (#547)
 │   ├── tools_analytics_registry.py        # mureo_analytics_modules_list / mureo_analytics_run (#440)
 │   ├── tools_creative_studio.py           # creative_studio_* (visual generation + compose)
 │   ├── tools_learning.py                  # mureo_learning_insights_get / mureo_consult_advisor
@@ -124,7 +127,13 @@ mureo/
 │   └── errors.py        # Context-specific errors
 ├── analysis/            # Analysis utilities
 │   ├── lp_analyzer.py   # Landing page analyzer
-│   └── anomaly_detector.py  # Zero-spend / CPA-spike / CTR-drop detection (pure, sample-size-gated)
+│   ├── anomaly_detector.py  # Zero-spend / CPA-spike / CTR-drop detection (pure, sample-size-gated)
+│   └── exclusion_impact/    # Delivery-impact preview for bulk exclusions (#547; pure)
+│       ├── models.py        # DeliveryRecord / ExclusionTarget / ExclusionImpact + coverage verdicts
+│       ├── matching.py      # Per-entity-kind match rules (host, app id, negative keyword match type)
+│       ├── estimator.py     # The share itself — incremental and cumulative
+│       ├── rules.py         # ## Guardrails exclusion keys + the one refusal decision
+│       └── surfaces.py      # Which tools are exclusion surfaces (mureo's + plugin-registered)
 ├── rollback/            # Rollback feature (allow-list gated, append-only audit trail)
 │   ├── models.py        # RollbackStatus enum + RollbackPlan dataclass
 │   ├── planner.py       # plan_rollback(ActionLogEntry) -> RollbackPlan | None
@@ -169,6 +178,7 @@ docs/integrations.md          # Platform discovery + external MCP integration gu
 | Keywords | `keywords.list`, `keywords.add`, `keywords.remove`, `keywords.suggest`, `keywords.diagnose`, `keywords.pause`, `keywords.audit`, `keywords.cross_adgroup_duplicates` |
 | Negative Keywords | `negative_keywords.list`, `negative_keywords.add`, `negative_keywords.remove`, `negative_keywords.add_to_ad_group`, `negative_keywords.suggest` |
 | Negative Placements | `negative_placements.list`, `negative_placements.add`, `negative_placements.remove` |
+| Placement Performance | (read used by `analysis_exclusion_impact_preview`: `group_placement_view` via `get_placement_performance`) |
 | Budget | `budget.get`, `budget.update`, `budget.create` |
 | Accounts | `accounts.list` |
 | Search Terms | `search_terms.report`, `search_terms.analyze` |
@@ -227,7 +237,7 @@ These families are not tied to a single ad platform. Tool names are the exact MC
 |--------|-------|
 | Analytics Registry (#440) | `mureo_analytics_modules_list`, `mureo_analytics_run` |
 | Rollback | `rollback_plan_get`, `rollback_apply` |
-| Analysis | `analysis_anomalies_check` |
+| Analysis | `analysis_anomalies_check`, `analysis_exclusion_impact_preview` |
 | Creative Studio | `creative_studio_providers_list`, `creative_studio_generate_visual`, `creative_studio_edit_visual`, `creative_studio_compose`, `creative_studio_brand_kit_get` |
 | Learning | `mureo_learning_insights_get`, `mureo_consult_advisor` |
 | Learning pre-flight (#548) | `mureo_learning_reset_preflight` |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,6 +29,7 @@ mureo/
 ├── google_ads/          # Google Ads API client (Mixin composition)
 │   ├── client.py        # GoogleAdsApiClient (main entry)
 │   ├── mappers.py       # Response mapping to structured dicts
+│   ├── _placement_mappers.py # Negative-placement + group_placement_view row mappers (#544/#547)
 │   ├── _ads.py          # AdsMixin (RSA create/update/status/list)
 │   ├── _ads_display.py  # DisplayAdsMixin (RDA create + RDAUploadError)
 │   ├── _keywords.py     # KeywordsMixin (add/remove/suggest/diagnose)
@@ -132,7 +133,7 @@ mureo/
 │       ├── models.py        # DeliveryRecord / ExclusionTarget / ExclusionImpact + coverage verdicts
 │       ├── matching.py      # Per-entity-kind match rules (host, app id, negative keyword match type)
 │       ├── estimator.py     # The share itself — incremental and cumulative
-│       ├── rules.py         # ## Guardrails exclusion keys + the one refusal decision
+│       ├── rules.py         # ## Guardrails exclusion keys, the one refusal decision, inert-rule reporting
 │       └── surfaces.py      # Which tools are exclusion surfaces (mureo's + plugin-registered)
 ├── rollback/            # Rollback feature (allow-list gated, append-only audit trail)
 │   ├── models.py        # RollbackStatus enum + RollbackPlan dataclass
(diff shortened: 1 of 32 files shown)


hyoshi:
## Review fixes (55b91a7)

**1. The cumulative cap was silently inert for ad-group-scoped writes — the incident's own scope.**

Reporting `null` with a reason was honest about the gap, but honesty about a gap is not the operator knowing to cover it. A rule that cannot fire now says so **at the moment it cannot fire**:

- `unevaluated_rules(impact, rules)` — a pure function over the same two values `evaluate_exclusion_impact` takes, so the two cannot disagree.
- Surfaced as a `NOT ENFORCED on this call:` line in the notice appended to the exclusion's own result, naming `max_delivery_share_removed_pct` as the backstop when that is the missing piece.
- Surfaced as an `unevaluated_rules` array in `analysis_exclusion_impact_preview`.
- Symmetric for `max_delivery_share_removed_pct` when coverage is `unknown` — the same defect on the Meta surface.
- `_mureo-strategy`'s SKILL.md now says **"do not write this one alone"**, with the reason (campaign-level exclusions are unreachable from an ad-group call's arguments; Google Ads exposes no ad-group-level negative keyword listing) and the backstop to pair it with.

**2. `_targets()` accepted an empty `entity_type`.** Both `value` and `entity_type` are now required and non-blank. Extended to `delivery_records` and `standing_exclusions` too — an `entity_type` typo on a delivery row inflates the denominator without ever the numerator, which under-reports the impact, the same defect class in the same function pair. `additionalProperties: false` catches a misspelled *key*; this catches a missing or blank *value*.

**3. `mappers.py` was 817 lines.** `map_negative_placement`, `map_placement_performance` and `_PLACEMENT_TYPE_TO_KIND` moved to `mureo/google_ads/_placement_mappers.py` — the same split `_placements.py` got from the client. 735 + 125 lines. No re-export shim: `_placements.py` imports from the new module directly, so there is no cycle to keep working around; the placement-mixin import inside `map_negative_placement` stays lazy for the one that remains. A test now pins both modules under the 800-line budget so the next addition cannot quietly re-cross it.

**Verification.** 10 injections, 10 caught (the original 7 re-run clean, plus: inert cumulative rule not reported → 3 tests; notice drops the inert-rule lines → 1; blank `entity_type` accepted → 4). Clean venv: **7895 passed, 8 skipped**. `ruff` / `black` / `mypy` clean.

PR #575, opened 2026-08-07
feat: import externally-made changes into action_log (#545)

### LINKED ISSUE #545: Import externally-made changes (platform UI edits) into action_log, so mureo's guarantees survive manual operation

hyoshi:
Closes #545.

Based on `feat/batch-revertible-unit` (#549) — it touches `action_log`'s model
and the rollback planner, both of which #549 also changes.

## The gap

Every guarantee mureo offers hangs off mureo having **made** the change. An
operator working in a platform's own UI — normal professional work, not misuse
— never goes through `StrategyPolicyGate`, never reaches `action_log`, never
gets an `observation_due`, and never appears in `/daily-check`'s evidence step.

The consequence is not a thin log but a blind one: **mureo could not tell
"nothing happened" from "something happened that I cannot see"**. In the
originating incident that gap cost ~6 days of failed diagnosis, because nothing
connected "delivery died" to "exclusions were added".

## What lands

`mureo_external_changes_import` polls each configured platform's change feed and
appends what mureo did not do. `/daily-check` runs it at step 2b, before it
diffs or diagnoses anything. Idempotent.

### 1. Observed is not performed

`ActionLogEntry` gains `origin` / `external_id` / `occurred_at` — optional,
emitted only when set, so an existing STATE.json parses unchanged and gains no
new key on the next write.

`plan_rollback` refuses **every** external entry, before any other check and
even when it carries a well-formed reversal hint. mureo never captured the
prior value, so a reversal built from such a hint is a fresh change dressed as a
restoration — and an imported entry is exactly where a forged hint is cheapest
to plant, since its content comes from outside mureo entirely. A batch mixing
mureo's own changes with imported ones therefore reports `partial` coverage
instead of promising a revert it could only half deliver.

### 2. The observation window anchors on the change

14 days from `occurred_at`, not from the import — a change made three weeks ago
lands **already past due** and is reviewed on the next run.

`metrics_at_action` is deliberately unset. mureo was not there; synthesising a
baseline from today's numbers would have `mureo_outcome_evaluate` score a delta
against a "before" that never existed.

### 3. mureo's own changes are not double-counted

Every change mureo dispatches also appears in the platform's feed. The feed's
attribution fields cannot separate the two — `user_email` is the same OAuth
identity whether the operator worked in the UI or through mureo, and
`client_type: GOOGLE_ADS_API` covers every API tool on the account. So the
discriminator is mureo's own log: same platform, same target identity, within
10 minutes.

**Which way it fails is deliberate.** Where identity is missing on either side
the match cannot be made and the change is imported as external. An over-import
shows the operator a change they may have made through mureo — visible, mildly
annoying, correctable. An over-attribution silently swallows a real UI edit,
which is the exact blindness this exists to remove.

### 4. Missing coverage is reported, never smoothed over

Every configured platform appears in the response. No feed ⇒
`change_import_unavailable_for_<platform>` (the same contract as
`analytics_not_available_for_<platform>`); a failed feed ⇒ `error`. **Neither
means "no changes"** — the absence of a change feed is not evidence of
innocence. `blind_spots` collects them so a reader does not have to notice an
*absence*.

`truncated: true` when a feed capped its response: `change_event` returns at
most 100 rows with no paging and retains ~30 days. History cannot be
reconstructed after the fact, only captured continuously — which is why the
first import's lookback is a deliberately short 7 days rather than 30.

## Per-platform coverage

Honest answer, not roadmap. "Not read by mureo" ≠ "the platform has no change
history" ≠ "nothing happened there".

| Platform | Change feed | Read by mureo today | Gap |
|---|---|---|---|
| Google Ads | `change_event` (GAQL) | **Yes** — built-in feed | 100-row cap, no paging; ~30-day retention; user-made changes only (automated bidding / policy / automated rules do not appear) |
| Meta Ads | Ad Account Activity (`/activities`) | **No** | The edge exists on Meta's side; **mureo ships no client for it**. Not "Meta has none" |
| Amazon Ads (bridge) | Vendor API | **No** | Can opt in via the ABI hook; does not today |
| Yahoo Search / Display, LINE, SmartNews (plugins) | Vendor change history | **No** | Same — plugin-side, via the ABI hook |
| TikTok Ads (hosted connector) | The connector's own tools | **No, and mureo cannot** | Outside mureo's data path entirely — no credentials, no dispatch. A skill records what it reads there via `mureo_state_action_log_append` with `origin: "external"` |
| Search Console / GA4 | n/a | n/a | Not ad-serving platforms |

## ABI hook and its effect on installed plugins

A **new** `ChangeFeedProvider` Protocol in a **new** `mureo.change_feeds`
entry-point group.

- **New Protocol**, because adding a method to an existing `runtime_checkable`
  one silently de-registers every published implementation (`ABI-stability` §4)
  — the wall #546 hit and avoided by splitting `DeliveryCollapseModule` out.
- **New group** rather than a second Protocol inside `mureo.analytics`, so a
  bridge that only wants to publish a change feed does not have to stub four
  analytics methods it will never implement.

**Effect on existing plugins: none.** No existing Protocol, group, or dataclass
field set changed. A plugin that has never heard of change import registers
nothing here, is simply absent, and its platform is reported as unavailable.
Pinned by `test_an_existing_analytics_plugin_is_not_a_change_feed`.

One invariant is enforced rather than documented: `external_id` without
`origin="external"` raises. An external id on a mureo-originated entry would
poison dedup — the next import would treat mureo's own action as already
imported.

## Tests

TDD; the RED run before implementation was 26 failed / 9 passed on
`tests/test_change_import.py`. 49 tests across two files pin the five
properties that make this worth anything, plus the row mapping and the tool's
honesty.

**Regression injection** — 10 defects introduced one at a time, all caught:

| Injected defect | Caught by |
|---|---|
| Drop the external-origin refusal in `plan_rollback` | `TestRollbackRefusesExternal` (2) |
| Never attribute a change to mureo | `test_mureos_own_change_is_attributed_not_imported` |
| Never recognise an already-imported change | `test_second_import_of_the_same_change_is_a_no_op` |
| Report a platform with no feed as `imported` | `TestHonestUnavailability` (2) |
| Anchor the observation window on the import | `test_observation_window_anchors_on_the_change_not_the_import` |
| Drop the platform namespace from `external_id` | `test_id_is_namespaced_by_platform` |
| Drop campaign identity from a Google row | `test_identity_is_extracted_from_resource_names` |
| Trust a feed's platform label | `test_a_change_labelled_for_another_platform_is_dropped` |
| Forget what this pass already wrote | `test_the_same_change_twice_in_one_page_lands_once` |
| Allow `external_id` on a mureo entry | `test_external_id_without_external_origin_is_refused` |

## Gates

- `ruff check .` — clean
- `black --check .` — clean (623 files)
- `mypy mureo --ignore-missing-imports` — clean (302 files)
- `python -m pytest` in a **clean venv** — **7877 passed, 8 skipped**. Tool
  count pin 208 → 209 (verified in the clean venv; the dev machine has plugins
  installed and reports 503).

## Docs

`docs/change-import.md` (new — per-platform coverage, dedup rationale, hook
walkthrough), `docs/ABI-stability.md` §4b + §1 + §6, `docs/mcp-server.md`,
`docs/strategy-context.md`, `docs/architecture.md`, `AGENTS.md`, `CHANGELOG.md`,
and both copies of `skills/daily-check` + `skills/_mureo-shared`.

## Known limits, stated rather than smoothed over

- **No daemon polling yet.** The importer runs from `/daily-check` and on
  demand. #545 argues the always-on service is the natural home, and it is —
  but that is a separate lifecycle/credentials change. Until then, a gap in the
  daily schedule is a gap in the record, and the docs say so.
- **A first import cannot backfill.** What has aged out of a feed is gone.
- **Attribution to mureo is approximate by construction**, and fails toward
  over-import. See §3 above.


hyoshi:
Review response pushed as `7f8779c`. The CRITICAL is untouched here — confirmed still reproducing against `stamp_batch` on the base branch, see the last section.

## HIGH 1 — over-attribution

A match now requires the **same kind of change** as well as platform, identity and window.

Kind comes from a small shared vocabulary — `status` / `budget` / `bid` / `criterion` / `ad` / `ad_group` / `campaign` — derived independently from each side:

- **Feed side**: `changed_fields` is consulted *before* `resource_type`. That ordering is load-bearing: Google reports a budget edit as `CAMPAIGN` + `changed_fields=["campaign_budget"]` at least as often as `CAMPAIGN_BUDGET`, and reading only the resource type would file the first one as a generic campaign edit — the same bucket a status toggle lands in.
- **Log side**: the tool name, matched verb-before-noun, so `google_ads_campaigns_update_status` reads as `status` and not `campaign`.

**Both sides must yield a kind and the two must be equal.** "Unknown matches anything" would restore identity-only behaviour for every action mureo cannot classify — the permissive direction, and permissive here means silently discarding an operator's edit.

A **definite** create-vs-remove disagreement additionally refutes a match (mureo removing a negative keyword while the operator adds one on the same ad group in the same minute). It only ever refutes, never confirms: a `*_update` tool may be an upsert, so requiring agreement would block far more true matches than it protects.

### Your reproduction

```
entry kind = 'status'    (google_ads_campaigns_update_status)
feed  kind = 'budget'    (CAMPAIGN_BUDGET / amount_micros)
verdict    = import      <- was attributed_to_mureo
```

### Attribution matrix — same target (campaign 111) throughout

| entry kind / feed kind | window | verdict |
|---|---|---|
| criterion / criterion | inside | `attributed_to_mureo` |
| criterion / criterion | outside | `import` |
| status / criterion | inside | `import` |
| status / criterion | outside | `import` |
| status / budget (your repro) | inside | `import` |
| status / status | inside | `attributed_to_mureo` |
| unknown entry kind (free text) | inside | `import` |
| criterion / criterion, opposite op | inside | `import` |

Only same-platform + same-kind + same-target + inside-window is absorbed.

### What it still cannot discriminate

Written into `docs/change-import.md` under *What this cannot discriminate, stated plainly*, rather than left as an unmet claim:

- **Two same-kind changes on the same target inside the window.** mureo raises a budget, the operator raises it again two minutes later. Nothing in the feed separates them; mureo attributes the second to itself. This is the residual over-attribution and it is real.
- **Free-text / unrecognised action names** yield no kind, so nothing attributes to them — mureo's own change re-imports as external. Costs an over-import, not a swallow.
- **Sub-kind detail** — a targeting change and a negative-keyword change are both `criterion`. Splitting further trades one failure direction for the other.

## HIGH 2 — BYOD reported as checked

`ChangeFeedResult` gains `unavailable_reason: str = ""` (additive with a default, so no published feed changes). A registered feed that could not answer for this account or mode sets it; the importer maps a non-empty value to `UNAVAILABLE`, which is what puts the platform in `blind_spots`.

The Google feed sets it in BYOD mode instead of returning an empty result.

On the test: you were right that the name was the claim and the assertion was weaker. It now asserts `unavailable_reason` (the field that actually routes the status, not prose an operator may not read), and a second test carries the signal end to end — `handle_external_changes_import` → `blind_spots == ["google_ads"]`.

## MEDIUM — imports and batches

`join_active_batch=False`, with the reasoning recorded at the call site and in the docs: a batch is what the operator *did*, an imported change is not, and joining dropped the whole batch to `partial` coverage over an unrelated edit polled while the batch happened to be open. Same rule as the executor's own `rollback_of` exclusion. Pinned by `TestImportsStayOutOfBatches`.

Also documented the window trade-off next to `ATTRIBUTION_WINDOW_MINUTES`: widening can only *add* attributions, and the ones it adds are the least certain — a hand edit further from mureo's action, hiding behind it, failing silently. Narrowing can only add over-imports, which the operator sees.

## CRITICAL — not fixed here, as instructed

Confirmed **still reproducing** against `origin/feat/batch-revertible-unit` (no `stamp_batch` fix has landed yet). Via `mureo_state_action_log_append` with a batch open:

```
origin=None external_id=None occurred_at=None
is_external=False  plan_rollback -> supported
```

Note `join_active_batch=False` incidentally takes the *importer* path out of `stamp_batch`, so that path is unaffected either way — but the hosted-connector append path above is not, and still needs the base fix. I will rebase and re-verify once it lands.

## Verification

**Regression injection** — 7 further defects, one at a time, all caught:

| Injected defect | Caught by |
|---|---|
| Identity-only matching (revert the kind check) | `TestNoDoubleCounting` — 5 failures |
| Unknown kind matches anything | `test_an_unclassifiable_action_does_not_absorb_anything` |
| Read `resource_type` only, ignore `changed_fields` | `test_mureos_own_change_is_attributed_not_imported` |
| Never refute on operation | `test_opposite_operations_on_the_same_target_are_not_absorbed` |
| Treat "did not look" as a quiet window | `test_a_feed_that_could_not_look_is_unavailable_not_imported` |
| BYOD back to a bare note | `test_byod_says_so_...` + `test_byod_reaches_the_blind_spots_list` |
| `join_active_batch=True` | `test_an_imported_change_does_not_join_an_open_batch` |

Three earlier attribution tests were passing for the wrong reason once kind matching landed (their fixtures had mismatched kinds, so identity was no longer the discriminator). They now use same-kind fixtures so each isolates the property it names.

**Gates**: `ruff` clean · `black --check` clean · `mypy mureo` clean (302 files) · clean venv `pytest` **7888 passed, 8 skipped, exit 0**.

hyoshi:
Rebased onto `2d5d5c8`; outstanding verification complete. Pushed as `16d4bd4`.

## CRITICAL — verified on the path that is actually at risk

You were right that testing the importer path would prove nothing: `join_active_batch=False` takes it out of `stamp_batch` entirely. The verification runs through the **hosted-connector path** — `mureo_state_action_log_append` with a batch open, where joining is the default — and is now three permanent tests rather than a one-off check (`TestProvenanceSurvivesBatchStamping`):

| Assertion | Result |
|---|---|
| Entry joins the batch (`batch_id == batch.batch_id`) | ✅ it really is the stamping path |
| `origin` / `external_id` / `occurred_at` all survive | ✅ |
| `is_external` stays `True` | ✅ |
| `plan_rollback` on a forged allow-listed hint | ✅ `NOT_SUPPORTED` |
| `plan_batch_rollback` coverage | ✅ `none`, `apply_order == ()` — reported as a gap, not as covered |

**They catch the original bug, not just its neighbourhood.** Reverting `stamp_batch` to the enumerated rebuild fails all three with the exact symptoms:

```
E  AssertionError: assert None == 'external'
E  AssertionError: assert <RollbackStatus.SUPPORTED> is <RollbackStatus.NOT_SUPPORTED>
E  AssertionError: assert 'full' == 'none'
```

(First attempt at that injection failed on a `NameError` instead — `2d5d5c8` moved the `ActionLogEntry` import under `TYPE_CHECKING`. Redone with the import restored so the enumerated rebuild actually executed.)

Your `dataclasses.fields`-derived round-trip guard did its job on contact: it refused the three new fields until `_ENTRY_FIELD_VALUES` carried them. Added, with a note that these are the fields the enumerated version actually dropped.

## The residual limitation, promoted

Agreed it was the wrong weight — one bullet among three, for the one case that fails in the expensive direction, and the exact mixed-operation shape of the incident.

`docs/change-import.md` now has **"Known limitation: a same-kind edit within 10 minutes of a mureo change is lost"** as its own section, led by the advice rather than the mechanism:

> **If you edit an entity by hand shortly after mureo changed the same thing on it, mureo will record your edit as its own and you will not be told.**
>
> **What to do:** after mureo changes an entity, wait out the 10-minute attribution window before editing *that same entity's same setting* by hand — or make the edit and then say so, so the run has a record that does not depend on the import.

It then says plainly that nothing in the feed can separate the two (`user_email` identical, `client_type: GOOGLE_ADS_API` covers everything, rows otherwise identical), so widening what mureo compares cannot fix it — only the operator's timing or their own note can. And a matrix showing how narrow it is, which is what makes the advice usable rather than a blanket caveat:

| Situation | Imported? |
|---|---|
| Same entity, same setting, within 10 min of a mureo change | **No — silently attributed to mureo** |
| Same entity, same setting, after 10 min | Yes |
| Same entity, *different* setting (budget vs status) | Yes |
| Different entity, any setting | Yes |
| Anything when mureo made no nearby change | Yes |

The same advice is in `_mureo-shared`, phrased for the moment it applies — when the agent has just made a change and the operator says they are about to adjust the same thing by hand, it should say so. An agent reading only the skill now gets it.

The two failures that go the *other* way (free-text action names, sub-kind coarseness) are kept but demoted to their own list, explicitly labelled as over-import rather than silent loss. `dedupe.py`'s module docstring carries a pointer with a warning for the next person tempted to close the gap by loosening the comparison — every loosening moves more changes into that bucket.

## Gates

`ruff` clean · `black --check` clean · `mypy mureo` clean (302 files) · clean venv `pytest` **7893 passed, 8 skipped, exit 0**.

hyoshi:
Identity-layer CRITICAL fixed in `121aaa5`. You were right about the shape of the bug: intersect-and-accept is "some slot agrees", not "no populated slot disagrees", and it inverted the stated bias exactly where it mattered most.

## The fix

`_identity_pairs` intersection is gone. Two rules replace it:

1. **No slot populated on both sides may disagree** — reject on any disagreement rather than accept on any agreement. This is repro 1: the shared `campaign_id` is not a licence to ignore the `entity_id` that disagrees.
2. **Both sides must name their target at the same specificity** (`entity_id` > `ad_id` > `campaign_id`). A target and its container are not a match; an asymmetry there is *unresolved*, and unresolved means import. This is repro 2.

I used your precedence ordering as given — and it turns out to be the same one mureo already documents as the canonical target (#524, `/daily-check` step 9, `plugin_semantics.extract_mutation_identity`), so "how specifically can this side name its target" now has one answer across the codebase rather than a private one in the deduper.

### One consequence worth flagging

Requiring equal specificity forces the feed side to speak mureo's identity language. The Google adapter was reporting *both* the ad group and the campaign for ad-level rows, which under rule 2 looks strictly more specific than mureo's own record of the same change — so **every ad-level mureo action would have re-imported as external on every run**. Safe direction, but daily noise.

So the adapter now names **one canonical target per row**: an ad-level row names the ad (parsed out of `changed_resource_name`, the only place the feed names the ad itself) and not its parent ad group. A row whose ad cannot be resolved names no sub-campaign target at all, rather than falling back to the ad group and claiming specificity it does not have. Same rule `extract_mutation_identity` already applies to plugin mutations.

## Identity matrix

Kind held constant (criterion/create both sides), so only *who* and *when* vary:

| campaign | entity | present on | window | verdict |
|---|---|---|---|---|
| same | same | both | inside | `attributed_to_mureo` |
| same | same | both | outside | `import` |
| same | **different** | both | inside | `import` ← repro 1 |
| **different** | same id | both | inside | `import` |
| different | different | both | inside | `import` |
| same | entity on **entry** only | entry | inside | `import` |
| same | entity on **change** only | change | inside | `import` |
| same | ad on **entry** only | entry | inside | `import` ← repro 2 |
| same | ad on **change** only | change | inside | `import` |
| same | campaign level only | neither | inside | `attributed_to_mureo` |
| same | campaign level only | neither | outside | `import` |
| no identity | no identity | neither | inside | `import` |

Only two cells attribute, and both are an exact same-specificity same-value target inside the window.

## Tests

Both reproductions, both asymmetry directions, the identity-less pair, and a nine-case parametrised matrix. You were right that neither repro had a test — the suite only covered *different campaign entirely*, which is precisely why this shipped.

**Regression injection**, 5 defects:

| Injected | Caught by |
|---|---|
| Restore intersection-accept (the shipped bug) | 9 failures |
| Never reject on populated-both-sides disagreement | 5 failures |
| Ignore specificity mismatch | 4 failures |
| Allow two identity-less sides to match | 1 failure |
| Adapter promotes the parent ad group again | 2 failures |

The empty-identity injection came back **STILL GREEN** on the first pass — my test used an empty entry against a populated change, which rule 2 already rejects, so it never reached the guard. The guard only protects the case where *both* sides are empty (no slot to disagree, specificity trivially equal, so kind and time would be the entire test). Added that test; the injection now fails. Reporting it because a green injection is the only signal that a test is decorative.

## Docs

`| Different entity, any setting, any time | Yes |` was false and is corrected. The table now enumerates sibling, broader and narrower targets separately — all imported — and states plainly which row an earlier version got wrong, since an operator who read it was told not to look.

New section *How "the same target" is decided* carries both rules and the one-canonical-target obligation for feed authors.

**`ATTRIBUTION_WINDOW_MINUTES` note re-checked** as you asked. The asymmetry argument holds and is unchanged, but it was describing the wrong gap: the window now multiplies only the documented same-entity/same-setting case, where before it multiplied every edit anywhere inside a campaign mureo had touched. That correction is in the note, flagged as relevant when reading an older account's history.

## Gates

`ruff` clean · `black --check` clean (623 files) · `mypy mureo` clean (302 files) · clean venv `pytest` **7909 passed, 8 skipped, exit 0**.

hyoshi:
All three fixed in `a6a9b7f`. Reporting each separately as asked.

---

## 1. The invented proto field — confirmed, and worse than one name

I verified against the vendored SDK rather than taking it on trust:

```
google-ads 29.2.0, v23 ChangeEvent descriptor:
  ad_group, asset, campaign, change_date_time, change_resource_name,
  change_resource_type, changed_fields, client_type, new_resource,
  old_resource, resource_change_operation, resource_name, user_email

"changed_resource_name" in proto -> False
hasattr(ChangeEvent(change_resource_name="x"), "changed_resource_name") -> False
```

You are right on both counts. It was in the GAQL SELECT (server would reject the query) **and** in `map_change_event`, where `getattr` yields `""` silently. So `google_ads_change_history_list` — shipping today — was likely inert in production, and my "name the ad, not its ad group" work never fired against real traffic.

**I audited the whole mapper field-by-field against the descriptor** as instructed, not just that name. All ten other reads are correct, and `changed_fields.paths` exists. That was the only invented one.

**Why it survived, and what now stops it.** Every test built row dicts by hand and bypassed the mapper; and `MagicMock` answers `hasattr` for anything and returns a truthy mock, so a misspelling passes every assertion in the existing test. Three tests close it:

- `test_reads_real_proto_field_names` — drives `map_change_event` with a real `ChangeEvent`; a protobuf raises on a name it does not have.
- `test_every_mapped_key_exists_on_the_proto` — every emitted key ⊆ descriptor fields. Derived from the proto, so it keeps working as the mapper grows.
- `test_every_selected_field_exists_on_the_proto` — the same for the **GAQL SELECT**, which nothing checked at all: `_search` is mocked in every test in that file, so the query string was never read by anything. My first injection run left this one **STILL GREEN**, which is how I found it.

---

## 2. `AD` rows — confirmed in-diff regression

`AdService` names ads `customers/<cid>/ads/<id>`: no ad-group segment, no `~`. My parser only knew the composite `adGroupAds/<ag>~<ad>` shape, so `AD` rows came back campaign-only — worse than before my commit, and it would have re-imported every creative edit as external on every run. Both shapes accepted now, and the parametrised test covers `AD` alongside `AD_GROUP_AD`.

---

## 3. The flagship scenario — fixed, but with a limit I am naming rather than hiding

Your diagnosis was right and I had claimed more than I delivered. Both sides needed criterion-level identity and neither had it.

- **Capture**: `criterion_id` / `criterionId` now precede `ad_group_id` in `_ENTITY_ID_KEYS` (first match wins, so ordering is the fix), yielding `entity_type="ad_group_criterion"`.
- **Feed**: criterion rows now name the criterion parsed from `change_resource_name`, not the parent ad group — `AD_GROUP_CRITERION` → `ad_group_criterion`, `CAMPAIGN_CRITERION` → `campaign_criterion`.

That second half is **one step beyond what you sanctioned**, and I want it visible: without it the entry says criterion-777 and the feed says ad-group-222, which disagree, so the operator's edit is safely imported — but mureo's own keyword work would then re-import as external on every run. It is ~30 lines in a file I already own, symmetric with the ad case, with tests. Say the word and I will drop it.

**What is still not closed**, and is now in the docs instead of being claimed away: mureo does not record native keyword / exclusion / creative mutations at all — only status toggles are automatic. Everything else depends on the agent calling `mureo_state_action_log_append`, and nothing enforces the granularity. `_mureo-shared` now tells it to pass the criterion id (and `ad_id`, not the ad group, for ads), but that is guidance, not a mechanism. So the docs row reads:

> A *sibling* entity in the same campaign — **Yes, provided the sibling was recorded at its own granularity; see below**

with a paragraph saying plainly which paths do and do not guarantee it. A proper fix is native-mutation auto-recording (an `action_log` equivalent of `native_reversal` covering non-status tools), which is a separate piece of work and does not belong in this PR.

---

## Also: `entity_type`

You were right that it was called identity and never compared. The generic slot now folds type and id together, so `ad_group:999` ≠ `ad_group_criterion:999`.

---

## Verification

Six injections, all caught (the GAQL one only after adding the test above):

| Injected | Caught by |
|---|---|
| Mapper reads `changed_resource_name` again | 2 mapper tests |
| GAQL selects `changed_resource_name` again | `test_every_selected_field_exists_on_the_proto` |
| `AD` rows lose their ad id | `test_an_ad_level_row_names_the_AD_not_its_ad_group[ad]` |
| Criterion rows fall back to the ad group | 2 criterion tests |
| `criterion_id` discarded at capture | 2 `TestCriterionIdentity` tests |
| `entity_type` not compared | `test_the_same_id_under_a_different_entity_type_is_not_the_same_target` |

**Gates**: `ruff` clean · `black --check` clean · `mypy mureo` clean (302 files) · clean venv `pytest` **7921 passed, 8 skipped, exit 0**.

hyoshi:
`28359d9`. **The sweep is clean — there is no second instance.** Detail below, since a verified negative is only worth what the method behind it is worth.

## 2. GAQL sweep — CLEAN

**56 queries across 16 files, 360 field references, zero invalid.**

| file | queries | fields |
|---|---:|---:|
| `mureo/adapters/google_ads/adapter.py` | 1 | 5 |
| `mureo/google_ads/_ads.py` | 4 | 30 |
| `mureo/google_ads/_ads_display.py` | 1 | 1 |
| `mureo/google_ads/_analysis_auction.py` | 3 | 23 |
| `mureo/google_ads/_analysis_keywords.py` | 2 | 19 |
| `mureo/google_ads/_analysis_rsa.py` | 1 | 9 |
| `mureo/google_ads/_creative.py` | 1 | 9 |
| `mureo/google_ads/_diagnostics.py` | 7 | 27 |
| `mureo/google_ads/_extensions_callouts.py` | 1 | 4 |
| `mureo/google_ads/_extensions_conversions.py` | 6 | 29 |
| `mureo/google_ads/_extensions_sitelinks.py` | 2 | 13 |
| `mureo/google_ads/_extensions_targeting.py` | 9 | 58 |
| `mureo/google_ads/_keywords.py` | 4 | 32 |
| `mureo/google_ads/_media.py` | 1 | 8 |
| `mureo/google_ads/accounts.py` | 2 | 7 |
| `mureo/google_ads/client.py` | 11 | 86 |
| **total** | **56** | **360** |

**Two corrections were needed before that result meant anything.** Both are now asserted rather than assumed, because I do not think you should take a clean sweep on trust:

1. **It first reported 18 failures, all false.** Every one was a GAQL `type` field. proto-plus mangles names colliding with Python builtins (`type` → `type_`) while GAQL uses the unmangled API name — so `asset.type` is correct and my checker was wrong. Had I reported those as findings you would have been chasing eighteen non-bugs.

2. **It initially missed queries built from implicitly concatenated string literals** — `"SELECT "` `"campaign.id, "` — because the tokens arrive with quotes attached. That silently skipped **27 field references across two files** while still printing "clean". A sweep that reports clean because it looked at nothing is worse than no sweep, so `test_extraction_covers_every_query` now pins the extracted query count against an independent count of `FROM` clauses (56 = 56), and `test_every_query_selects_at_least_one_field` catches a query whose fields all failed to parse.

**Mutation-tested, not just run.** Planting `campaign.nope_not_real` into a triple-quoted query, into a concatenated query, blinding the extractor regex, and suppressing field parsing all fail the suite. The sweep lives in `tests/test_gaql_field_names.py`, derived from the tree and the SDK rather than enumerated, so it keeps working as queries are added.

**Nothing further needs releasing.** The one inert tool is the one already fixed in this PR.

## 1. `AD_GROUP_BID_MODIFIER`

Added. You were right that "not reachable today" is not a reason — it uses the same composite `parentId~criterionId` shape, so leaving it out kept the exact ad-group collapse for one resource type.

I also verified every path segment the adapter parses against the SDK's own `*_path` builders:

```
ad_path                   -> customers/1/ads/999
ad_group_ad_path          -> customers/1/adGroupAds/222~999
ad_group_criterion_path   -> customers/1/adGroupCriteria/222~777
campaign_criterion_path   -> customers/1/campaignCriteria/111~777
ad_group_bid_modifier_path-> customers/1/adGroupBidModifiers/222~777
```

All five match. That is now a test asserting against the builders rather than hardcoded strings — a wrong segment yields "no identity" silently rather than raising, so it is the same class of failure as the fabricated field and deserves the same guard.

## 3. `plugin_semantics`

Left as-is, documented as a known property in the module docstring: the undeclared-identity fallback labels guesses with hardcoded Google-shaped `entity_type` values applied to third-party arguments that merely share a spelling. Inherent to guessing rather than made worse here, unreachable through mureo's own built-ins, and avoidable by declaring `identity` — which is the supported route.

## Gates

`ruff` clean · `black --check` clean · `mypy mureo` clean (302 files) · clean venv `pytest` **7926 passed, 8 skipped, exit 0**.

Issue #603, opened 2026-08-12
google_ads logs SDK exceptions with exc_info=True, which prints request metadata


hyoshi:
## What happens

`mureo/google_ads/` logs Google Ads SDK exceptions with `exc_info=True` in ~35 places. That writes `traceback.format_exception()` — which includes `str(exc)` — to whatever handler is attached.

`GoogleAdsException` does **not** curate its `__str__`. Its `__init__(self, error, call, failure, request_id)` never passes a message to `super().__init__()`, so `BaseException` keeps the raw constructor args and formatting the exception prints the underlying `grpc.Call` repr — which carries `debug_error_string()`, and with it the request metadata: the developer token and the `authorization` header.

## Why it was invisible until now

Nothing in the package installed a logging handler, so every one of those lines was discarded at runtime. #581 (PR #599) installed one, which is what turns them into bytes on disk.

The three sites reachable from the configure server's Google OAuth callback — `mureo/google_ads/accounts.py` in `list_accessible_accounts()` — were fixed in that PR, because installing the handler is what made them dangerous. This issue tracks the rest.

## Where the rest are

~35 sites across:

- `mureo/google_ads/_ads.py`
- `mureo/google_ads/_analysis_btob.py`
- `mureo/google_ads/_analysis_budget.py`
- `mureo/google_ads/_analysis_performance.py`
- `mureo/google_ads/_creative.py`
- `mureo/google_ads/_diagnostics.py`
- `mureo/google_ads/_extensions_sitelinks.py`
- `mureo/google_ads/_monitoring.py`

`grep -rn "exc_info=True" mureo/google_ads/` finds them.

These are on MCP tool paths rather than the configure server, so today they reach whatever handler the host process installs — nothing, for a stdio MCP server. That makes them lower urgency than the three already fixed, not safe: any operator or host that turns logging on gets the same exposure, and the configure handler's reach may grow.

## The pattern to apply

The codebase already answers this correctly everywhere else, with the reasoning written down:

- `mureo/cli/web_auth.py` (the Google account-listing caller) logs `type(exc).__name__` and explains why in a comment.
- `mureo/meta_ads/accounts.py` scrubs `str(exc)` and breaks the chain with `raise ... from None`, because an httpx `HTTPStatusError` embeds the request URL.
- `mureo/amazon_ads/lwa.py` uses `type(exc).__name__` / `from None` throughout.

[...]

PR #608, opened 2026-08-13
chore: release v0.10.45

hyoshi:
Cuts `[Unreleased]` to `0.10.45`.

## Contents

6 commits since v0.10.44. Most of this release is one chain: installing the first logging handler in the package turned a number of previously-dead `logger.*` lines into bytes on disk, and each fix uncovered the next.

**Security**

- The configure server's OAuth callback access lines carried the `?code=` authorization code; query strings are now redacted before logging (#581)
- Google Ads SDK exceptions were logged with `exc_info=True`. `GoogleAdsException` does not curate its `__str__`, so formatting it prints the underlying `grpc.Call` repr — carrying `debug_error_string` and with it the developer token and `authorization` header. Fixed at the three configure-reachable sites (#581) and the remaining 36 (#603), with an AST regression guard so it cannot come back
- `_extract_error_detail` — the function nominated as the safe alternative — returned `str(exc)` in its own fallback, and that value is embedded in the `RuntimeError` callers see. Now returns the class name (#603)
- The mureo home directory is created `0o700`. `Path.mkdir(parents=True, mode=...)` applies the mode to the leaf only, so a fresh install created it at the umask default (#581)

**Meta Ads**

- A token pasted into the advanced credentials form was handed to the refresh clock under the previous token's timestamp and overwritten on the next API call (#578)
- An expiring token now warns on the dashboard, with a re-authenticate control on the credential row that opens the paste card directly instead of restarting the setup wizard (#579)

**Reporting honesty**

- An authentication failure is a first-class outcome: platforms answer with a machine-readable `auth_error` envelope, and `/daily-check` marks the report partial and withholds recommendations that depend on the missing platform instead of shipping error text where numbers should be (#580)
- `unrecognized_key` and `duplicate_account` no longer make contradictory claims about the same platform key (#606)

**Operability**

- `mureo configure` writes a documented, rotating log at `~/.mureo/logs/configure.log`, level via `MUREO_LOG_LEVEL` (#581)

## This PR

- `CHANGELOG.md` — `[Unreleased]` cut to `[0.10.45] - 2026-08-13`
- `pyproject.toml`, `mureo/__init__.py`, `.claude-plugin/plugin.json`, `gemini-extension.json` — `0.10.44` → `0.10.45`
- 53 `SKILL.md` frontmatter `version` fields synced

No functional change. 58 files, +59/-57.

## After merge

Publishing a GitHub Release for `v0.10.45` triggers `release.yml`, which uploads to PyPI via trusted publishing.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "mureo",
-  "version": "0.10.44",
+  "version": "0.10.45",
   "description": "Your local-first AI ad ops crew for Google Ads, Meta Ads, Search Console & GA4. mureo sits on top of the official ad-platform MCPs, gates every change against your strategy, correlates outcomes locally, and keeps an auditable decision log — credentials never leave your machine.",
   "author": {
     "name": "Logly Inc.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## [Unreleased]
 
+## [0.10.45] - 2026-08-13
+
 ### Fixed
 
 - **Two conflict notes on the same platform card contradicted each other**
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -1,6 +1,6 @@
 {
   "name": "mureo",
-  "version": "0.10.44",
+  "version": "0.10.45",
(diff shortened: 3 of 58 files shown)


hyoshi:
Holding the release until #609 and #610 land, so they ship in the same version. The CHANGELOG cut will be redone then — reopening this one would split the release in two.

Issue #638, opened 2026-08-17
A stale figure is rendered as the selected window's headline number


hyoshi:
## What happens

A client card renders, as its headline figures:

> **25,862** cost   **2** conversions   **12,931** CPA
> *Stale figures — updated 11 days ago*

The client's actual spend for the window on screen was **0**. Delivery had stopped: the flight ended, and there were nine consecutive zero days. The card was reporting an 11-day-old number as this window's answer, with the disclosure rendered as a small badge beside it.

The operator read the bold number, not the badge, and reported the dashboard as wrong. They were right to.

## Why

Two behaviours compose into a false statement.

**The figure is not replaced when a platform stops answering.** `merge_metrics_into_state` is explicit:

> Platforms the digest did not collect are left entirely alone: absence of a number is not a zero.

That is correct for an API failure — writing 0 because a request timed out would be a lie. But the card cannot distinguish it from the other cases, and neither can the operator.

**The stale figure is still presented as the window's value.** `_platform_freshness` computes `stale` correctly (the window's own length plus one grace day — a sound rule), and the card then renders the number at full strength and demotes the staleness to a badge. The card asserts "this is the figure for the window you selected" and that assertion is false.

Compare how the same file treats a conflict: the client's totals are **withheld** until it is resolved, and the card shows `—`. Staleness is the same class of problem — mureo cannot vouch for the number — and gets the opposite treatment.

## Proposed fix

1. When `stale` is `True`, stop presenting the figure as the selected window's value. Show `—` for the headline metrics, as the conflict path already does, and surface the stale figure explicitly labelled with its age ("11 days ago: ¥25,862"). Nothing is hidden; it simply stops being stated as something it is not.

2. Record why a platform was not collected, so the card can say it. Today "not collected" and "collected, and the answer was zero" are indistinguishable in STATE.json — the worker's `platform_errors` never reaches the document. Without this an operator cannot act on a stale card, which is why this one sat for eleven days.

`stale is None` (unknown) keeps its current rendering — see #637 for why that state is so common today; the two issues are worth landing together, because a document with no `fetched_at` can never reach `stale=True` and would silently keep the old behaviour.

## Acceptance

- A rollup older than its window's stale threshold renders `—` for the headline metrics, with the stale figure and its age shown explicitly.
- A fresh rollup is unchanged.
- `stale is None` is unchanged.
- A platform the digest could not collect records why, and the card shows it.
- A regression test pins that a stale figure is never rendered in the position that asserts the selected window.

[...]

PR #641, opened 2026-08-17
feat: a platform can say why its figures did not move (#638)

hyoshi:
## Why

`Refs #638` — the second half of it. #639 stopped a stale rollup being rendered as the selected window's answer; the card now reads `—` and restates the stored figures with their age. That tells an operator the card is out of date. It does not tell them **why**, and without that there is nothing to do about it: an ad account that stopped delivering and a collector that stopped running produce an identical card. That is why the card in the report sat untouched for eleven days.

"Not collected" and "collected, and the answer was zero" were the same STATE.json. The worker's `platform_errors` never reached the document, so the fact existed and had nowhere to live.

## What this adds

**`platforms[<p>].not_collected`** — `{attempted_at, reason}`, optional and additive. A document written before it parses unchanged and gains no key on the next write; a cleared note leaves no key behind.

- **Model** (`mureo/context/models.py`) — defensive copy like every other mutable field, `None` by default, and a `NOT_COLLECTED_REASON_MAX_CHARS` bound the two boundaries share.
- **Codec** (`state_codec.py`) — declared in `_CODEC_COVERAGE`, parsed tolerantly (a hand-edited value costs the note, never the document), emitted only when present. **No reason means no note**: one that carries only a timestamp says something happened and refuses to say what, which is the non-answer this field exists to end.
- **Write path** (`state.py`) — `set_platform_not_collected(path, platform, account_id, *, reason)`. `attempted_at` is server-stamped (#460). Shares the platform-key guard with every other targeted writer.
- **Wire** (`web/reports.py`) — `_platform_row` carries `not_collected` beside `freshness`, whitelisted to the two known keys exactly as `totals` is, and `None` explicitly so every row has one shape.
- **Card** (`reports_logic.js` / `dashboard.js` / `i18n.json` / `app.css`) — the reason is rendered directly under the stale note it explains and above the repair hint, in en and ja. It also renders on the platform card, which is the only surface a single-client (OSS) install has.
- **MCP** — `mureo_state_platform_not_collected_set`, so a skill on a filesystem-less host can record and clear it.

## Two design points

**A note about the collection, never a verdict on the figures.** The stored `totals` / `periods` are left exactly as they were, because they are still the last numbers truly collected — writing `0` for a request that timed out would be the lie `merge_metrics_into_state` already avoids. Nothing here withholds or restates a number; that decision stays with staleness alone (a test pins that `aggregateClientKpis` never looks at this field). Both localized strings say it outright: *"The figures shown are the last ones collected — they are not wrong, they are older."*

**Clearing is explicit, and it belongs to whoever collects.** `reason=None` (omit `reason` on the tool) removes the note; nothing else does. An omitted field means *leave it alone* everywhere else in this document, and one window's rollup landing does not prove the platform-level collection recovered — so a successful `set_platform_metrics` deliberately preserves the note rather than guessing. The contract is stated in the model, the writer's docstring and the tool description, and a test pins that a clear is distinguishable from an omission. A note that outlived its failure would be the same defect one field over: stale information, stated with confidence.

Recording a failure also does **not** re-stamp `last_synced_at` — a collection that failed is not a sync, and re-stamping it would report the document as just-synced on the strength of nothing having been collected.

## Tests

- `tests/test_platform_not_collected.py` — the field, the writer, the wire and the tool. Includes the two that matter: clearing vs omitting, and `last_synced_at` untouched.
- `tests/js/reports_not_collected.test.js` — executes the shipped `reports_logic.js` bytes: which rows carry a note, which localized sentence states it, and that a summary without one produces nothing.
- `tests/test_web_assets_reports_not_collected.py` — positional pins: the reason follows the stale note, precedes the repair hint and the KPI cells, and never reaches the figures.
- `tests/test_dataclass_field_preservation.py` grew the field's fixture value (its guard fired on the new field, as designed) and a preservation test for the new mutator.
- `browser_contract.test.js`'s module table gained the three exports in both `bound` and `moved`.

## What is NOT in this PR

`Refs`, not `Closes`: the write side lives in mureo-agency. This PR gives the digest somewhere to put `platform_errors` and defines the contract it has to honour — record on failure, **clear on the next success** — but until that lands, no note is written by anything and the card is unchanged for every existing document. The issue's other acceptance criteria (stale figures withheld, fresh unchanged, `stale is None` unchanged, the positional regression test) landed in #639.


hyoshi:
Review response — `d355683`.

## CRITICAL — a fresh `fetched_at` and an older failure could be shown together

Fixed server-side, in `mureo/web/reports.py`:

```python
def _platform_not_collected(state: PlatformState) -> dict[str, Any] | None:
    note = _safe_not_collected(state.not_collected)
    if note is None:
        return None
    attempted = _parse_timestamp(note.get("attempted_at"))
    if attempted is None:
        return note
    collected = _newest_collection(state)
    if collected is not None and collected > attempted:
        return None
    return note
```

`_newest_collection(state)` takes the newest parseable `fetched_at` across `totals` **and every `periods` bucket**, through the existing `_parse_timestamp` (trailing `Z` tolerated) — no string comparison, and nothing duplicated into `reports_logic.js`. `_platform_row` now calls `_platform_not_collected(state)` rather than `_safe_not_collected(state.not_collected)`; the whitelist helper is unchanged and still does the key/length hygiene.

The three limits you named, each with a test:

- **Any window retires it.** The comparison is against the whole entry, not the window on screen, so the period toggle cannot resurrect a retired note (`test_the_period_toggle_cannot_resurrect_a_retired_note`).
- **No `fetched_at` anywhere keeps the note** — the never-collected case, which is the one the card has nothing else to say about (`test_a_platform_never_collected_at_all_keeps_its_note`, `test_an_advisory_platform_with_no_rollups_keeps_its_note`).
- **Retirement must be provable.** An unparseable `fetched_at`, or a note with no `attempted_at`, leaves the question open — open is not retired, the same position `_platform_freshness` takes (`test_an_undatable_pair_keeps_the_note`).

`test_the_note_never_touches_the_figures` is kept: it pins a different invariant (the note stays out of the KPI aggregate), which is still true and still worth pinning now that the contradiction is resolved a layer up.

## HIGH — the contract was silent on the tool called on success

`mureo_state_platform_metrics_set`'s description now says: *"If this platform carries a `not_collected` note (a previous collection failure), clear it in the same pass — call `mureo_state_platform_not_collected_set` with `reason` omitted; this call preserves the note rather than guessing that one window's rollup means the platform recovered."* Pinned by `test_the_metrics_tool_names_the_clearing_call`.

## MEDIUM 3 — a re-pointed key carried the old account's note

`_platform_base` drops `not_collected` when an entry that names an account is re-pointed at a *different* one, so every targeted writer inherits it. `""` → an id is stamping identity onto an entry that claimed none, so the note stays; `act_` spellings are one account (`account_ids_match`). Three tests.

## MEDIUM 4 — the contradiction itself was untested

`TestARetiredNoteIsNotShown` (7 tests) is exactly that: the arrangement where `fetched_at` is newer than `attempted_at`, plus every case that must NOT be suppressed. The two pre-existing tests were also made explicitly coherent — `test_the_note_rides_alongside_freshness_rather_than_replacing_it` now dates its note *after* the figures, so it tests the arrangement the note is true of rather than accidentally straddling both.

## LOW 5 — `reason` uncapped on read

`_parse_not_collected` now truncates to `NOT_COLLECTED_REASON_MAX_CHARS`, so the constant's "both boundaries" docstring is true at all three.

## Docs

`docs/strategy-context.md`, `skills/_mureo-strategy/SKILL.md` (+ mirror) and `CHANGELOG.md` all state the read rule and that the clearing contract is a duty rather than what correctness rests on.

## Verification

Red before green on the CRITICAL: `5 failed, 34 passed` — `test_a_collection_that_succeeded_since_retires_the_note`, `test_a_success_in_ANY_window_retires_it`, `test_the_period_toggle_cannot_resurrect_a_retired_note`, `test_repointing_a_key_at_another_account_drops_the_note`, `test_an_over_long_reason_is_truncated_on_read`.

Green after: `40 passed` for the suite; `418 passed` across the touched Python suites; `node --test tests/js/*.test.js` → 169 pass / 0 fail; `black --check` and `ruff check` clean; full run `5 failed, 9386 passed, 8 skipped, 37 deselected` — the 5 + the 9 deselected `live_clients` are the 14 that fail on this machine before the branch.

Issue #645, opened 2026-08-18
The duplicate block asks which entry holds the right figures without showing either entry's figures


hyoshi:
## What happens

The undecidable-duplicate block now names the command that ends the conflict (#636), and then asks for a judgement it does not give the operator the means to make:

```
One ad account (13097) is stored under platform keys that
BOTH name real platforms (logly_ads_context, plugin:mureo-logly-bridge), so its spend, conversions and CPA are
counted twice.
mureo does not choose between them: the two entries usually hold different
partial figures, so dropping either under-counts as much as adding them
together over-counts. Decide which entry holds the right figures, then
remove the other by naming it:
 mureo repair platform-key --key <the key to remove> --drop-duplicate
```

**"Decide which entry holds the right figures" — and neither entry's figures are shown.** The block prints the account id and the two keys, and stops.

## Why this is the same shape as #636

The module docstring opens with who this command is for:

> The person who needs this cannot read JSON and cannot judge which of two entries holds the right figures, so the command is built around what it can decide safely and shows its work for everything else

and the first bullet promises exactly what is missing here:

> **A dry run is the default.** … It names the keys, the ad account and what each entry holds (`fetched_at`, which windows, whether a `totals` rollup is there)

That promise is kept for entries in the plan — `_repair_preview._totals_line` / `_periods_line` print the window, the fetch time, the rollups, the campaign count. It is not kept for the one finding the command explicitly hands back to the operator, which is the only finding where the operator has to weigh the entries themselves.

The operator's way out today is to guess that running the drop command twice — once per key, without `--apply` — will show what each entry holds. Nothing on screen says so. #636 was closed on the principle that an instruction needs a runnable next step; this is an instruction whose next step is a judgement with no evidence on screen.

Observed on a real 26-client sweep, where an operator who is not an engineer was told to decide and could not.

## Proposed fix

Print each entry's facts under the duplicate block, in the same shape the planned-repair preview already uses — reuse `_totals_line` / `_periods_line` / the campaign count / `_echo_conversion_override` rather than writing a second wording of the same facts.

`PlatformEntryFacts` for both sides is already available: `_describe` builds it from the document, and `undecidable_groups` already has the group. The data is there; only the printing is missing.

Worth stating explicitly in the block that the two dry runs are safe, or making them unnecessary by printing the facts up front. The latter is better: an operator comparing two entries should see them side by side, not in two separate command outputs they have to hold in their head.

## Acceptance

[...]

Issue #643, opened 2026-08-18
An entry holding only a not_collected note is treated as an empty stub


hyoshi:
## What happens

`PlatformEntryFacts.is_empty_stub` (`mureo/context/platform_repair.py`) does not know about `not_collected`, which #638 added in v0.10.48. Its own docstring states the criterion:

> Does this entry store nothing a removal could lose?

For an entry whose only content is a collection-failure note, that answer is now wrong — removing it loses the note.

`_describe` never reads `not_collected`, so the fact cannot reach the property:

```python
return not (
    self.campaign_count
    or self.has_totals
    or self.rollups
    or self.conversion_action_types
)
```

## Why an entry can hold only a note

A platform that failed on its FIRST collection has no campaigns, no `totals` and no `periods` — the note is all there is. Both writers deliberately create such an entry, because that is the case an operator can least diagnose: `set_platform_metrics`' sibling `set_platform_not_collected` in this repo, and `merge_metrics_into_state` in mureo-agency (logly/mureo-agency#205).

## When it bites

`_drop_reason` reaches `DROP_EMPTY_STUB` only for a key that does not resolve, so a note-only entry is safe wherever its plugin is installed. It is not safe elsewhere — and "elsewhere" is not exotic:

- the bridge is uninstalled, or a client is opened on a colleague's machine;
- `mureo repair platform-key --all` is run there.

Resolvability is a fact about the machine, not about the entry — that is the whole finding of #609 / #631. So the entry that best explains why a platform has no figures is dropped exactly on the machines where the operator cannot see the platform either, and dropped silently, as "empty".

## Why the note belongs with `conversion_action_types`

The repair's stated licence to drop figures is that figures come back:

> The figures are not lost — they are re-fetchable from the platform, and the pre-repair document is backed up first.

`conversion_action_types` is excluded from dropping for the opposite reason: no sync restores it (#617). **A `not_collected` note has that same property.** It records that a collection failed at a stated time; nothing re-derives it. A later run either succeeds — retiring the note through the read-side rule — or fails again and writes a NEW note about a new attempt. The record of the earlier failure is gone.

So the note falls on the `conversion_action_types` side of the line the module already draws, and `is_empty_stub` should count it.

## Proposed fix

Carry `not_collected` into `PlatformEntryFacts` (via `_describe`) and count it in `is_empty_stub`, so an entry holding only a note is not an empty stub.

[...]

Issue #652, opened 2026-08-19
No code path reads the monthly budget target, so pacing cannot reach any surface the budget-pacing skill is not already running on


hyoshi:
## What happens

`## Custom: Monthly Budget` is already the established home for an operator's monthly spend target. `skills/budget-pacing/SKILL.md:30` states its standing:

> **`## Custom: Monthly Budget`** section in STRATEGY.md, if present — the operator's explicit monthly figure (and optional per-platform sub-targets). This wins; it is the intended monthly spend, not a safety ceiling.

and `:32` already has the skill offer to persist it when it is missing, so an agent asked for a monthly budget writes it back in that shape.

**No code reads it.** The section is reachable only by an agent that loads the skill, which means monthly pacing — is this client over or under its intended spend — cannot appear anywhere the model is not already running that skill. It cannot appear on the Reports view.

## Why this is the gap and not a missing feature

The reading machinery exists and is already used for the neighbouring section. `mureo/policy/strategy_gate.py:403`:

```python
def guardrails_from_strategy_text(text: str) -> Guardrails:
    for entry in parse_strategy(text):
        if entry.title.strip().lower() == GUARDRAILS_HEADING:
            return parse_guardrails(entry.content)
    return Guardrails()
```

`mureo/context/strategy.py` supplies `parse_strategy` / `render_strategy` / `add_strategy_entry` / `write_strategy_file`. So STRATEGY.md is already a structured document that code reads and writes — this is not prose parsing, and this issue does not introduce a new coupling. It adds the twin of a function that has been in the tree for a while.

## Guardrails is not the right home, and this is the important distinction

`Guardrails` carries ceilings — `max_daily_budget_per_campaign`, `max_total_daily_budget`, `max_daily_budget_increase_pct` — values whose job is to **refuse an operation that exceeds them**.

A monthly budget target is not a ceiling. **Underspending it is a problem too**, and no operation should be blocked for approaching it. The skill says this outright (`:31`): a daily ceiling multiplied by days in the month is an *implied cap*, to be labelled as such and used only when no explicit target exists.

Putting the target into `Guardrails` would make an intended figure look enforceable, which is the opposite of what it is.

## Suggested fix

A reader for `## Custom: Monthly Budget` alongside `guardrails_from_strategy_text`, returning the total and any per-platform sub-targets, and **"not set" as a first-class answer** rather than a zero or an inferred value.

Two properties matter more than the parsing:

[...]

Issue #651, opened 2026-08-19
The multi-client Reports view has no triage layer, so a client that needs attention is not distinguishable from one that does not


hyoshi:
## What happens

An operator running many clients cannot tell, from the Reports view, which client needs attention today. The view is a flat grid of client cards; everything mureo knows about a client that is *wrong* is rendered inside that client's own card, at the same visual weight as everything that is fine.

Two field incidents this week show the cost, and both were visible on screen the whole time:

- A client's card carried a double-counted ad account (#636). The card was red and its totals were withheld, and it stayed that way until someone happened to look.
- A client's card showed an eleven-day-old figure as the selected window's spend (#638). Delivery had stopped; nobody noticed for eleven days.

Neither was a missing signal. Both were **unsurfaced** signals.

## Why this is worth building now

Everything the triage layer needs already exists and is already computed per client — this issue is aggregation and ranking, not new data:

- platform-key conflicts that withhold a client's totals (`_build_platform_conflicts`, #636)
- stale rollups, with the age (`_platform_freshness`, #638)
- why a platform was not collected (`not_collected`, #638)
- observations due (`action_log`, surfaced today by `judge_client_health`)

## Scope: the multi-client view only

The Reports client grid is rendered by OSS (`mureo/_data/web/dashboard.js`) but populated through the Agency seam (`mureo/web/report_clients.py::_agency_list_clients`, which returns `None` when Agency is absent).

**The triage layer must appear only where that seam supplies clients.** A single-workspace install has no other client to triage against, and its per-card signals are already in place. Do not degrade this to a one-row version for single-workspace — omit it.

## Requirements

1. **Rank by what mureo can act on.** A conflict that withholds totals outranks a stale figure, which outranks an observation due. State the ordering in code, do not leave it to render order.
2. **Every item says what to do next.** #636 was reported precisely because the dashboard said "resolve this" and no command could. An item with no next step is a bug in the item, not a display detail.
3. **A withheld figure is a triage item, not a blank.** This is the constraint that matters most. #638 established that mureo does not present a number it cannot vouch for; the triage layer must therefore have a first-class state for "we cannot say", never an empty cell that reads as zero or as fine.
4. **Silence when there is nothing.** No "0 alerts" banner competing for attention with the cards.
5. Counts must agree with the cards below. If the layer says three clients need attention, exactly three cards are marked.

## Out of scope

- Monthly-budget pacing alerts — needs the budget target to be readable by code (separate issue).
[...]

PR #653, opened 2026-08-19
chore: release v0.11.0

hyoshi:
## Release v0.11.0

Version bump and CHANGELOG cut only. No functional change is in this PR — every
diff line is a version string, plus the `[Unreleased]` → `[0.11.0] - 2026-08-20`
heading.

### Why minor, not patch

#648 adds a **new public contribution point for third-party plugins**:
`mureo.policy.platform_model.register_platform_model`, together with the
`PlatformModel` dataclass, `Evidence` and `PlatformModelWarning`.
`docs/ABI-stability.md` lists it as a stable, additive surface — a package
outside this repo can now depend on it. That is a feature addition, so a patch
bump would understate it.

### What is in this release

**Added**

- **A platform plugin can now state how its own platform works, on the one
  route that is always read** (#648). Until now the only text a plugin could
  put in front of the model unconditionally was its MCP tool names and
  descriptions; a `SKILL.md` contributed through `mureo.skills` is
  description-matched, so its body is read only when the agent decides the
  description applies — which does not happen on a daily-check or
  weekly-report run. A plugin could therefore ship a correct account of its
  platform and have it never be read at the point where a borrowed mental
  model does its damage: the report itself.

  `register_platform_model` is that missing channel. A provider registers one
  capped paragraph — how its platform selects and prices delivery, and what it
  therefore does *not* have — and mureo renders it into the MCP server's
  `instructions`, which the client receives inside the `initialize` response:
  before any tool call, with no description to match.

  The guards are the point as much as the channel is. Every model carries an
  `Evidence` record (first-party source, ISO retrieval date, the sentence it
  rests on) and is refused at registration if any of the three is missing — a
  plausible sentence with no source is indistinguishable from a correct one
  until it costs money. Registration is **first-wins** and a model renders only
  where a tool matching its `tool_prefix` was contributed by the provider the
  model names, so a plugin may speak for itself and never for another platform.
  Statements are capped at 400 characters and the rendered block at 2000, with
  dropped platforms named to the operator and flagged to the agent rather than
  silently truncated. mureo core ships **zero** built-in models: a platform
  with no registered model contributes no text at all, and the default install
  still gets a byte-identical `InitializeResult`.

**Fixed**

- **Auction and smart-bidding semantics were stated as unconditional facts**
  (#647). mureo's platform-agnostic instruction text and its shared entity
  vocabulary described bidding, auctions and smart-bidding learning phases as
  properties of advertising rather than of the platforms that work that way. On
  a platform whose delivery is not selected by a bid those sentences are simply
  false, and the agent carried them over anyway — into routine report lines and
  into stored learning entries, which outlive the session that produced them
  because nothing downstream re-derives whether the premise held. The
  unqualified statements are now qualified the same way the already-correct
  ones in `budget-pacing` and `daily-check` were. `BidStrategy` gained a
  `NOT_APPLICABLE` member (additive, non-breaking per `docs/ABI-stability.md`
  §5) so that "this platform has no bid strategy" is distinct from `None`,
  which already meant *unknown / not fetched*.

- **A platform entry holding only a "why this was not collected" note was
  deleted as empty** (#643). `mureo repair platform-key` drops an entry on the
  proof that it stores nothing a removal could lose, but `not_collected` (#638)
  was not counted — so an entry whose only content was the note read as nothing
  at all. Figures come back on the next sync; a note does not. The note now
  sits on the `conversion_action_types` side of that line and appears in the
  dry run beside the campaign count, the `totals` window and the stored
  periods.

- **The block asking which duplicate to keep showed neither entry's figures**
  (#645). When one ad account is stored under two keys that both name real
  platforms, mureo refuses to choose and asks the operator to decide — but
  printed only the account id and the two key names, so an operator who is not
  an engineer was told to decide with nothing on screen to decide from. Both
  entries are now described in the block that asks, through the same helpers
  every other block uses.

### Verification

- `ruff check .` — All checks passed
- `black --check .` — 720 files unchanged
- `mypy mureo/ --ignore-missing-imports` — no issues in 349 source files
- `node --test tests/js/*.test.js` — 169 pass, 0 fail
- `pytest tests/ -q` — 9495 passed, 8 skipped, 14 failed. All 14 failures are
  pre-existing on the machine this was prepared on and are unrelated to the
  version bump (locally installed companion packages alter the MCP tool list
  and the analytics client registry). CI is the authority here.

### Not done in this PR

Merging, tagging and publishing the GitHub Release.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "mureo",
-  "version": "0.10.48",
+  "version": "0.11.0",
   "description": "Your local-first AI ad ops crew for Google Ads, Meta Ads, Search Console & GA4. mureo sits on top of the official ad-platform MCPs, gates every change against your strategy, correlates outcomes locally, and keeps an auditable decision log — credentials never leave your machine.",
   "author": {
     "name": "Logly Inc.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## [Unreleased]
 
+## [0.11.0] - 2026-08-20
+
 ### Added
 
 - **A platform plugin can now state how its own platform works, on the one
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -1,6 +1,6 @@
 {
   "name": "mureo",
-  "version": "0.10.48",
+  "version": "0.11.0",
   "description": "Your local-first AI ad ops crew. Works with Claude Code, Cursor, Codex & Gemini.",
   "contextFileName": "CONTEXT.md"
 }
--- a/mureo/__init__.py
+++ b/mureo/__init__.py
@@ -1,3 +1,3 @@
 """mureo — your local-first AI ad ops crew. Works with Claude Code, Cursor, Codex & Gemini."""
 
-__version__ = "0.10.48"
+__version__ = "0.11.0"
--- a/mureo/_data/skills/_mureo-amazon-ads/SKILL.md
+++ b/mureo/_data/skills/_mureo-amazon-ads/SKILL.md
@@ -2,7 +2,7 @@
 name: _mureo-amazon-ads
 description: "Amazon Ads (official MCP, bridged by mureo): query campaigns, ad groups, ads and targets, run reports, and manage account access under Amazon's own tool names."
 metadata:
-  version: 0.10.48
+  version: 0.11.0
   openclaw:
     category: "advertising"
     requires:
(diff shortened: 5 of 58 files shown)


hyoshi:
Closing without merging — not abandoned.

The release is held until the Reports dashboard work lands (#651, #652). Keeping this branch open until then would mean re-resolving a CHANGELOG conflict on every dashboard merge and re-editing the cut section each time, which is churn that buys nothing and adds chances to get it wrong.

The cut will be re-made from `main` once that work is in. Nothing here needs to be preserved: the file set is derived mechanically from the previous release commit, not from this branch.

Issue #223, opened 2026-06-12
Dashboard plugin-credentials cards render twice: concurrent renderPluginCredentials() races (clear→await→append) during double renderAll() at init


hyoshi:
## Symptom

The dashboard "Plugin credentials" section shows **every provider card twice** (e.g. LINE Ads / Yahoo! JAPAN Ads ×2). Reported on an agency install; reproduced with browser instrumentation.

## Measured reproduction

On dashboard load (`#dashboard`):

```
resource-timing entries for /api/credentials/plugins : 2   ← double fetch at init
provider cards rendered                              : 4   ← 2 unique × 2
after one MUREO.setLocale() switch                   : +1 fetch → 2 cards (correct)
```

## Root cause

`renderPluginCredentials()` is async: it clears the container, **awaits** the fetch, then appends. During dashboard init, `renderAll()` runs **twice** (two init paths invoke it), so two render calls overlap:

```
A: clear → await …      B: clear → await …
A: append N cards       B: append N cards   ⇒ 2N cards
```

The per-call `container.textContent = ""` cannot help once the calls interleave. A locale switch later runs a single render and "fixes" the page — which is why the duplication looks locale-related (it reappears whenever two renders race again).

## Proposed fix

Make the render generation-safe (covers any number of concurrent triggers, init or otherwise):

```js
let pluginRenderSeq = 0;
async function renderPluginCredentials() {
  const seq = ++pluginRenderSeq;
  …fetch…
  if (seq !== pluginRenderSeq) return;   // a newer render superseded us
  container.textContent = "";            // clear *after* the await, right before append
  …append…
}
```

(Deduplicating the double `renderAll()` at init is also worth doing, but the guard makes the section robust regardless.)

## Related

Reported as logly/mureo-agency#71 (initially suspected agency code; the agency removed its own plugin-credentials card in mureo-agency#62, so the only renderer of this section is dashboard.js).

PR #709, opened 2026-08-26
Roster usability: the whole row opens its client, and the remembered view gets a reload test

### LINKED ISSUE #707: Roster usability: the table's rows are not clickable, and the table/cards choice is reported not to come back
Two things an operator hits on the agency roster table (#691 phase 3), reported from daily use.

## 1. Only the "Detail →" cell opens a client

The table's rows are 44px tall and carry the client's name, health dot and
every figure — the whole row reads as one object — but the only part of it
that opens the client is the button in the last column. Every other table in
this dashboard's world (and the card grid this view replaces) opens on the
whole target, so an operator aims at the row, clicks, and nothing happens.

What it should do:

- A click anywhere on the row opens that client's detail view.
- Not at the cost of anything else in the row: a click that lands on a real
  control (`a` / `button` / `input`) is that control's, not the row's, and a
  click that ends a text selection is a selection, not a navigation.
- Keyboard parity: the row is focusable and Enter opens it.
- `cursor: pointer` and a hover tint, so the row looks like what it now is.
- The explicit "Detail →" button stays. It is the affordance a screen reader
  announces and the one an operator can point at; a whole-row target is an
  addition to it, not a replacement.

## 2. Reported: the table/cards choice does not come back

Phase 3 remembers the roster view in `localStorage` under
`mureo.reports.roster_view`, per browser. The report from daily use is that
the chosen view is not the one showing on the next visit.

To be established in the fix: whether the stored key round-trips, whether the
restore happens before the first draw, whether the "a roster that shrank to
one client keeps the cards" rule can fire for a roster that did not shrink,
and whether a refused write is being swallowed. Whatever it turns out to be,
the regression test this needs is the one nothing covers today: choose a
view, reload, and assert the choice is what draws — with the storage carried
across the two page loads rather than seeded into one.

hyoshi:
Closes #707.

## 1. The whole row opens its client

The table's rows are 44px tall and carry the client's name, its health dot
and every figure — the row reads as one object — but the only part of it
that opened anything was the button in the last column. A click anywhere on
the row now opens that client, and the row carries `tabindex="0"` so Enter
does the same.

It takes only the clicks nobody else wanted:

- **A click on a real control is that control's.** `a`, `button`, `input`,
  `select`, `textarea`, and anything `contenteditable` that is not `"false"`,
  checked from the event target up to the row. So the "Detail →" button still
  opens the client **exactly once** — the test asserts the count, not just
  the slug — and a link added to a cell later still leads where it leads.
- **A click that ends a text selection is a selection.** Dragging across two
  cells to copy a figure ends in a click, and a row that navigated on it
  would take the screen away at the moment the operator finished reading,
  losing the copied figure with it.
- **The row's table semantics stay.** No `role` is set: a `<tr>` that claimed
  to be a button would stop being a row and the figures under it would stop
  being announced with their columns. The row is merely focusable, and the
  announced control for a client is still the "Detail →" button — which is
  why it is kept rather than replaced.

`cursor: pointer` and a `--surface-2` tint on hover say the row is a target.
A tint rather than the card's `border-color` change, because these rows share
collapsed borders and a border that lit up would move the row below it; the
focus outline is inset for the same reason.

Row and button now share one `openRosterClient()`, so a client opened either
way lands in the same place with the same selection recorded.

## 2. The remembered table/cards choice — what was measured

The report was that the chosen view is not the one showing next time. **It
did not reproduce.** What was measured, against the bytes that ship, driven
through the real `app.html` + `app.css` in `dom_harness`:

| Hypothesis | Result |
|---|---|
| (a) the write and the read disagree on the key | **No.** One `ROSTER_VIEW_KEY` const serves both, and the new round-trip test fails when they are made to differ. |
| (b) the restore runs after the default is drawn | **No.** Both containers ship `hidden`, and `rosterViewFor()` runs inside `renderRoster()` before anything is drawn — there is no `await` between `setReportsView("index")` and the first draw, so no frame in between. |
| (c) "a shrunk roster keeps the cards" misfires | **No.** `rosterViewFor()` only forces cards below `ROSTER_TABLE_MIN`; checked at 2 and 3 visible clients, and with an archived client in the roster. |
| (d) a refused write is swallowed | **Real, but not this.** `writeRosterView` does swallow a `setItem` that throws — deliberately, and documented in the file — so a browser that refuses storage gets a toggle that silently forgets. That is not what a normal Chrome on a loopback origin does. |

Also ruled out, each by direct measurement rather than by reading: a stale
installed wheel (the file in site-packages is byte-identical to `main`, at
0.15.0); a duplicate `<script>` tag; a second copy of the assets in
`mureo-agency` / `mureo-pro`; a key collision with `mureo.locale`,
`mureo.reports.client_order` or `mureo.reports.triage.dismissed`; the
`[data-reports-view]` selector also matching the `data-reports-view-switch`
wrapper (it does not — attribute selectors match the exact name); the
double-`renderAll()` at boot (#223) racing a toggle click; a click landing
mid-render; navigating away and back; and a period switch. Every read path
re-reads storage, so the preference cannot be lost while the store persists.

**So the PR does not carry a fix for it — it carries the test that was
missing, and the finding.** Every test covering this preference seeded the
key into a page and asserted what drew; none let one page WRITE the choice
and a second page READ it. `openRoster({ carry })` copies a finished page's
store into the next one, exactly as a browser carries `localStorage` across
F5. Asserted in both directions — "table" is also the default, so a test that
only ever asserted the table would pass with the storage unread — plus across
a client detail and back, which is the other "next time" an operator means.

The one explanation left standing is **origin instability**, and it is worth
checking before changing any code: `localStorage` is scoped to
`scheme://host:port`, and the configure server prefers 7613 but falls back to
an **ephemeral port** when it is busy (`ConfigureWizard._bind_server`, logged
at INFO and never surfaced; the service status readers exist precisely to
"prefer the actually-bound port/url from configure.json so an ephemeral-port
fallback is reported"). A different port is a different, empty store. So is
`localhost:7613` versus `127.0.0.1:7613`. If that is what happened, the
language would have reset to English and the card drag order would be gone
**at the same time** — one question settles it.

## Verification

- `node --test tests/js/*.test.js` — **515 pass, 0 fail** (was 498; 17 new).
- Mutation-checked, since green tests only prove the code passes its tests:
  dropping the control guard fails 3, the selection guard 1, `tabindex` 1,
  accepting a key other than Enter 1, opening the wrong slug 3; making the
  write key differ from the read key fails 4, and neutering the restore
  fails 3.
- `pytest tests/` — 14 failed / 9908 passed on this branch and **the same 14
  on `main`** (`582e061`); diff of the two runs is empty but for the elapsed
  time. All 14 are pre-existing in this dev environment (locally installed
  plugins change the tool inventory).
- `black --check` clean over 744 files; `ruff check` clean.

No new i18n strings, no new served asset, no Python touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 ## [Unreleased]
 
+### Changed
+
+- **A roster row opens its client from anywhere on the row** (#707). The
+  agency table's rows are 44px tall and carry the client's name, its health
+  dot and every figure — the row reads as one object — but the only part of
+  it that opened anything was the button in the last column. A click anywhere
+  on the row now opens that client, and the row is focusable so Enter does
+  the same.
+
+  It takes only the clicks nobody else wanted. A click that lands on a real
+  control (`a` / `button` / `input` / `select` / `textarea`, or anything
+  `contenteditable`) is that control's, so the "Detail →" button still opens
+  the client exactly once and a link added to a cell later still leads where
+  it leads. A click that ends a text selection is a selection: dragging
+  across two cells to copy a figure ends in a click, and a row that navigated
+  on it would take the screen away at the moment the operator finished
+  reading, losing the copied figure with it.
+
+  No `role` is set on the row. A `<tr>` that claimed to be a button would
+  stop being a row and the figures under it would stop being announced with
+  their columns, so the row is merely focusable and the announced control for
+  a client is still the "Detail →" button — which is why it stays.
+
+### Fixed
+
+- **The remembered table/cards choice now has a reload test behind it**
+  (#707). The roster view is stored per browser under
+  `mureo.reports.roster_view`, and the tests covering it seeded that key into
+  a page rather than letting one page write it and the next read it. A key
+  the two halves disagreed about, or a restore that ran after the default was
+  drawn, would have passed every one of them. The round trip is now asserted
+  in both directions, plus across a client detail and back.
+
 ## [0.15.0] - 2026-08-26
 
 ### Changed
--- a/mureo/_data/web/app.css
+++ b/mureo/_data/web/app.css
@@ -4060,6 +4060,27 @@ dialog[data-confirm-dialog] [data-confirm-ok] {
   background: var(--status-ok);
 }
 
+/* The whole row opens its client (#707), so it has to look like it does.
+   A tint rather than the card's border-color change: these rows share
+   collapsed borders with their neighbours, so a border that lit up would
+   move the row below it. NO `display` here — `.roster tbody tr[hidden]`
+   above is what takes a filtered row off screen, and nothing may outrank
+   it. */
+.roster-row {
+  cursor: pointer;
+}
+
+.roster-row:hover {
+  background: var(--surface-2);
+}
+
+/* Inset, because an outline drawn outside a collapsed-border row is clipped
+   by its neighbours on the two long edges. */
+.roster-row:focus-visible {
+  outline: 2px solid var(--accent-ring);
+  outline-offset: -2px;
+}
+
 /* A row that needs acting on carries the same left rule its card does, so
    the two views mark the same clients the same way. */
 .roster-row.is-attention td:first-child {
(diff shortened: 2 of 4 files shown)


hyoshi:
reopening to re-trigger CI (only CodeQL fired)

PR #721, opened 2026-08-28
Split dashboard_reports.js, and grade each client once per render

### LINKED ISSUE #715: Split dashboard_reports.js and precompute per-client health once per render
## What

`mureo/_data/web/dashboard_reports.js` is ~909 lines (guideline: ≤800, typical 200–400), and `renderReportsIndex` inside it is ~100 lines (guideline: ≤50). Both predate #713; #713 added ~13 lines of hero wiring and was reviewed with this flagged as a follow-up, not a blocker.

## Work

1. **Split the file** the way 3-a/3-b already split their halves (`reports_*.js` pure model + `dashboard_reports_*.js` DOM): extract the index-composition block of `renderReportsIndex` into a helper (e.g. `buildReportsIndexModel(rows, summaries)` returning `{triage, healthCounts, hero, portfolio}`) and move it out of the main file.
2. **Precompute per-client health once per render.** `triageClientHealth(triage, i)` is an O(items) scan and is currently recomputed in 4–5 places per render per client (inside `triageHealthCounts`, the hero's `healthOf` closure, the card loop, the roster renderer). Correctness-safe (pure function — no "two answers" risk) but a needless O(4·C·R) multiplier; compute a `healthByIndex` array once in `renderReportsIndex` and pass it everywhere.

Behaviour-preserving refactor only: every existing test stays green unmodified except pure import-path pins.

Refs #706, #713 (review MEDIUMs).

hyoshi:
Closes #715.

`dashboard_reports.js` was 909 lines and `renderReportsIndex` was ~100 of them. The function was doing two jobs at once: deciding what the list screen SAYS about a roster, and putting it on screen. This separates them.

## The split

Along the seam 3-a/3-b already established — `reports_*.js` is the decision, `dashboard_reports_*.js` is the DOM:

- **`reports_index.js`** (new, 139 lines) — `buildReportsIndexModel(rows, summaries)` returns `{triage, healthByIndex, healthCounts, hero, portfolio}`: the alert layer, the per-client health, the health split, the band and the portfolio figures. It decides nothing itself; it composes `reports_triage.js`, `reports_overview.js` and `reports_hero.js`, read off the page at call time exactly as `reports_hero.js` reads `reports_overview.js`.
- **`dashboard_reports_changes.js`** (new, 243 lines) — the detail view's tier (2), "what moved since yesterday, and what is wrong". Moved whole and byte-identical. It is about ONE client's two adjacent days, which is the question the index file does not ask; it was the block in that file answering something different from everything around it.

## One verdict per client, reached once

`triageClientHealth` is a scan of the alert layer, and the index used to run it four times per client per render: inside the health counts, inside the band's `healthOf` closure, on each card, and again on each roster row. Four scans that can only ever agree, because the function is pure — the multiplier was the whole cost, and on the twenty-seven-client screen it is the render's hot loop.

The model now holds `healthByIndex`, the verdict in grid order, and every consumer is handed the array.

**It is not a second opinion.** The verdict is still `reports_triage.js`'s. `triageClientHealth` fills the array; `triageHealthCounts` still counts it, taking the array as an optional third argument rather than rebuilding it. An entry counts only when it is one of the verdicts that function produces, and anything else falls back to asking it — so `triageHealthCounts(built, total)` with no array behaves exactly as it always did, which is what `dashboard_reports_overview.js` still calls.

## Behaviour

Preserved throughout: the same nodes appended in the same order, the same payloads, no i18n key added or dropped, no CSS touched. The pure builders were already pure, so hoisting them above the renderers changes nothing but the order two side-effect-free calls happen in.

| | before | after |
|---|---|---|
| `dashboard_reports.js` | 909 | 756 |
| `reports_index.js` | — | 139 |
| `dashboard_reports_changes.js` | — | 243 |
| `renderReportsIndex` | ~100 | 24 |

`renderReportsIndex` now reads as the five steps it is; `renderReportsIndexBands` (16) draws everything above and beside the grid, `renderReportsIndexGrid` (44) draws the grid, `reportsPeriodUnion` (9) is the window union it ends on.

## Tests

- **JS suite: 627 passing, up from 620, with no assertion changed.** The seven new cases are the ones the existing tables generate for a new module: `reports_index.js`'s row in `browser_contract.test.js`'s `MODULES` (it publishes one global, the same object Node sees, no CommonJS leak, `dashboard.js` names it when missing) and the parse check every served asset gets. The only other test edits are wiring: the two new files in `dom_harness.js`'s load order.
- **Python asset pins: 379 passing.** Eleven pins named `renderReportsIndex` and now name the function the line they were protecting moved to (`renderReportsIndexGrid`, `renderReportsIndexBands`, or `buildReportsIndexModel` in `reports_index.js`). Every claim they make is unchanged — "the layer is built in >1 place", "a filter never survives a re-render", "the badges are the layer's own verdict" — and two of them got stronger: `dashboard_reports.js` is now asserted to contain no `buildReportsTriage(` / `buildReportsPortfolio(` at all.
- `pytest -k web`: 1915 passed, 2 failed — the same two `test_web_handlers.py` failures present on `main` before this branch, untouched by it.
- `ruff` and `black` clean.

## Not in scope

`renderReportsSummary` (112 lines) and `renderReportsPeriodToggle` (64, mostly the comment block explaining the one-window rule) are both over the guideline and both predate #715, which named `renderReportsIndex` specifically. `renderReportsPortfolio` still calls `triageHealthCounts` itself — a fifth scan #715 counts but does not list among the consumers to convert, since it would mean changing a DOM module's signature. Both left for a follow-up.


hyoshi:
Review round, both findings addressed in `d26f901`.

## The model had no test that called it (HIGH)

`reports_index.js` shipped with substring pins and the module table's surface checks — both of which a `buildReportsIndexModel` returning constants would pass. Its siblings all have a suite that runs them; this one did not.

`tests/js/reports_index.test.js`, 13 cases in three groups:

- **It is the triage layer's answers, not a copy.** Nothing is asserted against a literal health. Every case compares the model against the same lower-level function run independently — `triageClientHealth` per client, `triageHealthCounts` over the grid, `buildReportsHero`, `buildReportsPortfolio`, `buildReportsTriage`. A literal would pass just as happily if the model quietly stopped consulting the layer, which is the exact failure the file exists for. The health case also asserts the fixture grades more than one verdict, or the comparison would hold for any constant.
- **A payload an older daemon could send.** A roster that is not an array, summaries that are not an array, a summaries array shorter than its roster (a failed fetch), and one longer than it. The grid is the roster, not the fetches, so none of these may shorten `healthByIndex`.
- **`triageHealthCounts` is handed the array, never trusted blindly.** The third argument is a shortcut, not a channel for a new vocabulary: an entry outside the layer's own verdicts, a `null`, an array of the wrong length, and a non-array all fall back to the decision function and produce the two-argument answer exactly.

## The strip was still counting for itself (MEDIUM)

Fair, and the PR body overclaimed. `renderReportsPortfolio` called `triageHealthCounts(triage, portfolio.total)` — a fifth scan per render, and worse, it made "the band, the chips and the strip are one answer" a claim about two objects that merely agreed.

New signature: **`renderReportsPortfolio(portfolio, triage, counts)`**, called as `renderReportsPortfolio(model.portfolio, model.triage, model.healthCounts)`. The strip's own call is gone, and so is the now-unused `triageHealthCounts` binding at the top of `dashboard_reports_overview.js`. `portfolio.total` **is** `rows.length` (`buildReportsPortfolio` returns `total: rows.length` over the same array), so not a figure on screen moves. A missing `counts` prints zeros rather than recomputing — a fallback that counted again is precisely what this removed.

`buildReportsIndexModel` is now the only call to `triageHealthCounts` in the shipped assets, and the claim is asserted rather than asserted-about-one-file: the hero pin walks every `mureo/_data/web/*.js` and requires zero occurrences of `triageHealthCounts(` outside `reports_index.js` and the definition in `reports_triage.js`. `AGENTS.md` and the `reports_index.js` / `reports_hero.js` docstrings now say five scans, not four, and name the strip.

`reports_hero_list.test.js` gains the behavioural half — the strip's health note asserted against the band's own blocks, so a dropped hand-off shows up as a strip saying "0 need attention" over a grid with a red card on it.

## Mutation check

Each mutation applied to the shipped bytes, suites run, then reverted:

| Mutation | Caught by |
|---|---|
| `healthByIndex` filled with `"ok"` instead of `triageClientHealth` | **20 JS tests** across 8 suites (`reports_index`, `reports_filter_interaction`, `reports_triage`, `reports_hero_list`, `reports_roster_table`) + 1 Python pin |
| card loop uses `"ok"` instead of `model.healthByIndex[i]` | **10 JS tests** (cards, marks, filter counts, roster) — **no Python pin**, because the substring survives in the roster closure. The behavioural suite is what holds this one. |
| third argument dropped from `renderReportsPortfolio` | 1 JS test (`agrees with the portfolio strip`) + 1 Python pin |
| `triageHealthCounts` trusts `known[i]` without checking the vocabulary | 2 JS tests in the new fallback group |

## Verification

- JS: **641 pass, 0 fail** (was 627; +13 new file, +1 strip agreement).
- Python asset pins: 379 pass. `pytest -k web`: 1915 passed, same 2 pre-existing `test_web_handlers.py` failures as `main`.
- `ruff` + `black` clean across `tests/` and `mureo/`.
- CI green on `d26f901`, all 13 checks.

Issue #690, opened 2026-08-21
Daily per-platform history in STATE.json: the document forgets yesterday's yesterday


hyoshi:
## What

STATE.json keeps exactly one rollup per canonical window (`YESTERDAY` / `LAST_7_DAYS` / `LAST_30_DAYS`), and every daily collection **overwrites** it (`set_platform_metrics` merges per window key, same key replaced — `mureo/context/state.py:667-680`). The value it replaces is gone. Consequently nothing in the product can answer "was yesterday better than the day before?" — no day-over-day delta, no trend line. The reports UI redesign needs both.

The data itself is already fetched: every platform family ships a daily delivery report (`google_ads/_analysis_performance.py:702` `get_daily_delivery_report(days=60)`, `meta_ads/_insights.py:200`, both BYOD clients), and daily-check's delivery-collapse step already pulls those rows — then discards them.

## Proposal (schema + writer + reader, following the `periods` precedent)

1. **Schema**: `PlatformState.daily: dict[str, dict] | None` — date-keyed rollup map, `{"2026-08-20": {spend, impressions, clicks, conversions, fetched_at}, ...}`. Same shape and merge rules as `periods`, with `YYYY-MM-DD` keys instead of window tokens. `None` default: old documents parse unchanged, the key never appears unless written (byte-stability preserved).
2. **Writer**: `set_platform_daily(path, platform, account_id, *, days)` modelled on `set_platform_metrics` — per-date-key merge (same date replaced idempotently, past dates preserved), `_stamp_fetched_at` per bucket, `guard_platform_entry_write`. Validation is shape-based, not vocabulary-based: `^\d{4}-\d{2}-\d{2}$`, and reject server-today-or-later (partial days must not be written — same reasoning as `delivery_collapse.py:32-37`). Retention cap applied at write: keep the most recent **35** days (28-day baseline + margin, matching `DEFAULT_BASELINE_DAYS`), drop older keys.
3. **A missing day stays missing.** No zero-filling — "not collected" and "zero" are different facts (the `not_collected` rule). Readers render gaps as gaps.
4. **MCP tool** `mureo_state_platform_daily_set` next to `mureo_state_platform_metrics_set`; date keys via `propertyNames` pattern since they cannot be enumerated. Tool count pins (`tests/test_doc_claims.py`, README ×2, docs ×2, `test_mcp_server.py`) all move to 225.
[...]

PR #722, opened 2026-08-28
The daily-history write, usable without a path and anchored to the account's day

### LINKED ISSUE #710: Make the daily-history write path usable by document-level writers: public retention trim, caller-anchored day
## What

The daily-history write path (#690) turned out to be hard for a downstream writer (the agency digest, agency#214) to use as shipped. It worked around both gaps; the workarounds are the evidence the API needs two fixes.

**1. The retention trim is private.** `set_platform_daily` is path-based and single-purpose, so a writer that merges `daily` inside its own atomic document write (the agency digest, which must land the rollup and its day together through a tenant-scoped store) cannot use it — and then has no public way to apply the 35-day retention rule. The digest now imports `mureo.context.state._capped_daily` directly: a private name whose upstream rename breaks a paying customer's nightly write. Expose the trim — either a public `capped_platform_daily(daily) -> dict` or a document-level `with_platform_daily(doc, platform, account_id, days) -> StateDocument` that guards, merges and trims without touching the filesystem (the path-based mutator would then wrap it).

**2. Completeness is judged by the host clock, not the account's day.** `set_platform_daily` rejects "server-today-or-later" dates using `mureo.core.clock.server_now` — the host's clock. Ad accounts close their day in the account timezone (Asia/Tokyo for every current client). On a UTC host between 00:00 and 09:00 JST — exactly when a nightly cron digest runs — yesterday-JST is still "today" in UTC and a genuinely complete day is refused. The check needs an explicit timezone/date anchor (e.g. an `as_of_date` parameter the caller resolves in the account's timezone), keeping the "no partial days" rule while letting the caller state whose day it is.

Both fixes are additive; the existing signature keeps working.

hyoshi:
Closes #710.

The day-grain write shipped in #690 as one path-based mutator. A writer that must land `daily` **together with other fields in one atomic document write** could not use it, and worked around both gaps it hit. The workarounds are the evidence for these two fixes. Both are additive — every existing call keeps working unchanged.

## 1. The merge is available without the file

New module `mureo/context/daily.py`:

| Public name | What it is |
| --- | --- |
| `with_platform_daily(doc, platform, account_id, days, *, as_of_date=None) -> StateDocument` | The whole daily write **minus the filesystem**: the platform-key guard, the per-date-key merge, the `fetched_at` stamping and the 35-day trim. Returns a new document; mutates nothing. |
| `capped_platform_daily(daily) -> dict` | The retention rule on its own, for a writer that has already merged its own map. |
| `DAILY_RETENTION_DAYS` | Moved here with the trim it belongs to; still re-exported from `mureo.context.state`. |

`set_platform_daily` is now a thin wrapper: it keeps the lock, the atomic write and the pre-lock refusal, and delegates the merge. **Neither route can drift from the other** — a test asserts the two produce the same document.

`mureo.context.state._capped_daily` stays as an alias of the public name. A downstream nightly write imports it today; removing it would make the fix the break it was meant to prevent.

## 2. Completeness is measured against the account's day, when the caller says so

Only complete PAST days are stored, and "today" was the **host's**. An ad account closes its day in the **account's** timezone. On a UTC host at 02:00 Asia/Tokyo — exactly when a nightly cron digest runs — yesterday-in-Tokyo is still today in UTC, so a genuinely complete day was refused for the whole 00:00–09:00 JST window.

Against `main`, with the clock at 16:30 UTC on 2026-08-20 (= 01:30 JST on the 21st):

```
daily module:      ABSENT -> No module named 'mureo.context.daily'
no-anchor write:   REFUSED -> daily key '2026-08-20' is not a complete day yet (server today is 2026-08-20)
anchored write:    TypeError -> set_platform_daily() got an unexpected keyword argument 'as_of_date'
```

The optional `as_of_date` (today, resolved in the account's timezone) is what the check measures against when supplied; omitted, it is the server clock exactly as before. A `datetime` is narrowed to its date, since resolving "now" in a timezone is what produces one; anything that is not a date is refused rather than silently ignored.

**The rule itself does not move.** A day at or after the anchor is still refused — an anchor states *whose* today it is, it does not buy a partial day. The refusal now names the anchor it used ("as_of_date is …" / "server today is …"), because two clocks can say "today" and a message that does not say which one sends the caller hunting the wrong one.

### The MCP tool takes it too

`mureo_state_platform_daily_set` gains an optional `as_of_date`, shape-pinned to `YYYY-MM-DD` by a schema `pattern` so the dispatcher refuses a free-text "today" before any handler runs (#660). The handler then parses it, because the pattern cannot tell `2026-02-30` from a real date and an anchor mureo cannot place must not fall back to the host clock silently. Tool count is unchanged (226).

## Tests

**+32** in `tests/test_platform_daily_history.py` (50 → 82 in that file), including:

- the reproduction — UTC host × Asia/Tokyo account × the small hours — as **two** tests: the day refused without an anchor (the defect, kept as a pin on the unchanged default) and accepted with one;
- a pin on the *scenario*, so it cannot quietly stop straddling midnight and leave the pair passing for no reason;
- the anchor's own day and any later day still refused, with and without an anchor;
- the end-to-end path through the real dispatcher (`handle_call_tool`) with schema validation and gates, plus the schema-level refusals;
- the document-level writer: no filesystem touched, input document unmodified, per-date merge, other platforms and every other field preserved, guards and trim shared, and the same document as the path-based route;
- the public trim: cap, order, an undatable key kept and uncounted, no mutation of the caller's map, and `_capped_daily` still resolving to it.

## Verification

- `pytest tests/` — **14 failed, 10175 passed, 8 skipped**; `main` at the same commit gives **14 failed, 10142 passed, 8 skipped** with the *identical* 14 names (pre-existing environment failures: `tests/analytics/builtin/test_live_clients.py` ×9, the tool-count/plugin-wiring trio, `test_web_handlers` ×2). Baseline-diff zero. `+33` = the 32 new tests plus the auto-generated `test_imports` case for the new module.
- `ruff check .` clean, `black --check .` clean (750 files), `mypy mureo/ --ignore-missing-imports` clean (362 files).

## Docs

`docs/strategy-context.md` (the anchor, and the document-level API), `docs/mcp-server.md` (tool table), `AGENTS.md` (architecture tree), `CHANGELOG.md` (Unreleased).


hyoshi:
## Review response — CRITICAL: an unbounded `as_of_date` switched the completeness rule off

Fixed in 72b9a17. The reproduction was real: `as_of_date` was taken on trust, so a far-future anchor made every date this side of it a "complete past day", on both the Python API and the MCP route.

**The anchor is now bounded against the one clock mureo trusts.** An `as_of_date` more than `_MAX_ANCHOR_DAYS_AHEAD` (**2**) days ahead of `clock.server_now().date()` is refused. Two days is the widest gap real timezones can open: civil offsets run UTC-12 to UTC+14, a 26-hour spread, so an instant that is date D in one place can be D+2 in another — 22:00–24:00 on D at UTC-12 is 00:00–02:00 on D+2 at UTC+14. The JST-account / UTC-host case this parameter was added for only ever needs +1; the bound is stated for every timezone rather than for the one in front of us. The reasoning is in the constant's own comment.

**No bound in the other direction, on purpose.** A past anchor only makes the check *stricter* — it can refuse a day, never admit one — so it is self-limiting, and a caller whose clock is behind learns it from an explicit refusal naming both dates rather than from figures quietly dropped. Stated in the `_completeness_anchor` docstring and tested.

**The refusal names both dates**, because the caller stated one and mureo compared it against another.

### Docs (the MEDIUM)

`mureo/context/daily.py` module + function docstrings, `set_platform_daily`'s docstring, the MCP tool's `as_of_date` description ("an `as_of_date` more than 2 days ahead of the server's own date is refused outright"), `docs/strategy-context.md`, `docs/mcp-server.md`, `CHANGELOG.md`.

### Tests (+9, file 82 → 91)

- the boundary **pair on the same call**: `+2` accepted, `+3` refused;
- the bound pinned as a **value** (`_MAX_ANCHOR_DAYS_AHEAD == 2`), so widening it is a decision rather than a refactor;
- the reproduction **negated on both routes** — anchor `+60` with rows dated `+30`, through `set_platform_daily` and through `handle_call_tool` (schema validation and gates included: the anchor passes the `YYYY-MM-DD` pattern, so nothing upstream of the write stops it), each asserting no STATE.json is created;
- `with_platform_daily` bounded too — a downstream writer calls it directly;
- both dates named in the message;
- a past anchor left alone (accepts an old day against an old anchor, still refuses a later one).

### Mutation check

| Mutation | Result |
| --- | --- |
| bound removed (`if False:`) | **5 fail** — `test_one_day_past_the_widest_gap_is_refused`, `test_a_far_future_anchor_cannot_file_days_nobody_has_lived`, `test_the_document_level_writer_is_bounded_too`, `test_the_refusal_names_both_dates`, `test_a_far_future_anchor_is_refused_through_the_real_dispatcher` |
| bound widened to 3 | **2 fail** — `test_the_bound_is_two_days`, `test_one_day_past_the_widest_gap_is_refused` |

`daily.py` was restored byte-identical after each (verified with `diff`).

### Verification

`pytest tests/`: **14 failed, 10184 passed, 8 skipped** — the identical 14 pre-existing environment failures `main` gives (which passes 10142). Baseline diff zero. `ruff check .`, `black --check .` (750 files) and `mypy mureo/ --ignore-missing-imports` (362 files) all clean.

